PR #74, opened 2012-12-14
Removing conflict markers inherited from SVN by accident.

gvwilson:
The boot camp FAQ had some SVN conflict markers in it, which tripped up the web site compiler.

### DIFF
--- a/bootcamps/faq.html
+++ b/bootcamps/faq.html
@@ -57,14 +57,9 @@ delivery. Our experience is consistent with recent research: students
 learn best in a <em>blended</em> environment that combines directed
 in-person instruction with self-directed online learning. We have
 therefore borrowed the boot camp format used so successfully by the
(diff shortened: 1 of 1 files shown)


jiffyclub:
Nice job, Greg, you've jumped right into the deep end! Unfortunately I made almost this exact change this morning in 231538eb2722, except I kept the other part of the conflict. Shall we just leave things as they are, or would you like to get intimate with git and try to reverse my changes, then apply yours?

gvwilson:
1. So why didn't I see this?  Did you not accept your own pull request 
   until just now?
2. Um... please take care of it?  And I'll get brave later?  Because I'm 
   not yet?

jiffyclub:
I must have made my update after you forked. We'll just leave things as is for now. It looks like the only difference in the conflict is a new line and we can add that back in later if we notice it.

PR #82, opened 2012-12-14
Updating Makefile to build into local 'build' directory by default, or into main web site if told to

gvwilson:


caitsydney:
Does the README.md file need to be update? Cause it says to use "make site" and I don't think that is an option anymore.

gvwilson:
Good catch --- fixed.

PR #92, opened 2012-12-15
Make the footer mailto link a label to match the GitHub link

jiffyclub:
I also moved the GitHub link over to the right side along with the mailto
link. I think that has better balance. The GitHub link looked pretty lonely
dwarfed by the CC badge.

### DIFF
--- a/_base.html
+++ b/_base.html
@@ -126,9 +126,17 @@
       </div>
 
       <div class="footer">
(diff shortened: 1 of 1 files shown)


jiffyclub:
Here's a preview:

Screen Shot 2012-12-14 at 10 45 40 PM

gvwilson:
OK, I can live with that (though it'd look even nicer if they were stacked).
G

jiffyclub:
I'm going to disagree, I can't find an elegant way to stack them:

Screen Shot 2012-12-14 at 11 05 52 PM

Issue #40, opened 2012-12-10
Create "check my machine" script for Windows


gvwilson:
Write a small script that people can download from our web site and run on their machine that will tell them if everything they need for a boot camp has been installed correctly --- see https://github.com/swcarpentry/website/issues/38 for starting points.

wking:
I expect my “Linux” script will work on MS Windows too, but I don't
have a Windows box available for testing.  Can someone with a Windows
box check the following and report back with the output:

  $ git clone -b installation-test --single-branch git://github.com/wking/swc-website.git
  $ cd swc-website
  $ python swc-installation-test-1.py
  Passed
  $ python swc-installation-test-2.py
  check virtual-shell...  pass
  check virtual-editor... pass
  check virtual-browser...        pass
  …

If something's broken, I can try and help fix them remotely, but it
would be nice if someone with some Windows and Python experience could
take the lead on compatibly patches.  I'll help keep you honest on the
POSIX side ;).

ian-mitchell:
I'm running cygwin.  The current version of git in cygwin is 1.7.9, so I don't have the --single-branch option.  However, after removing that option, I was able to clone the whole repository and get the correct branch.

I've got EPD 7.3-2 installed (so I can run ipython and load scipy, numpy, etc), but I'm guessing that my cygwin installation (where I ran these test scripts) cannot see the EPD installation.

```
mitchell@armstrong ~/Workshops/SWCarpentry/CheckMyMachine
$ git clone -b installation-test git://github.com/wking/swc-website.git         
Cloning into 'swc-website'...
remote: Counting objects: 3401, done.
remote: Compressing objects: 100% (2529/2529), done.
remote: Total 3401 (delta 840), reused 3362 (delta 819)
Receiving objects: 100% (3401/3401), 57.17 MiB | 781 KiB/s, done.
Resolving deltas: 100% (840/840), done.

mitchell@armstrong ~/Workshops/SWCarpentry/CheckMyMachine
$ cd swc-website/

mitchell@armstrong ~/Workshops/SWCarpentry/CheckMyMachine/swc-website
$ python swc-installation-test-1.py
Passed

mitchell@armstrong ~/Workshops/SWCarpentry/CheckMyMachine/swc-website
$ python swc-installation-test-2.py
check virtual-shell...  pass
check virtual-editor... fail
check virtual-browser...        fail
check Git (git)...      pass
check Mercurial (hg)... pass
[...]

Issue #39, opened 2012-12-10
Create "check my machine" script for Mac OS X


gvwilson:
Write a small script that people can download from our web site and run on their machine that will tell them if everything they need for a boot camp has been installed correctly --- see https://github.com/swcarpentry/website/issues/38 for starting points.

wking:
I just tested my “Linux” scripts on 10.6.8 and they work there too.  I
don't think we need different scripts for OS X.  If we get to the
point where we want to provide installation suggestions (instead of
just telling the user what they have installed), it will be easy
enough to switch on `os.name` to select the appropriate hint.

gvwilson:
Agreed: should we use 'posix' and 'win32' as names (since that's what 
os.name returns)?
Thx,
G

wking:
On Thu, Jan 03, 2013 at 06:47:19AM -0800, Greg Wilson wrote:

> Agreed: should we use 'posix' and 'win32' as names (since that's what 
> os.name returns)?

I'm fine with that, but we could also use `platform.system().lower()`
for marketing-style names [1](http://docs.python.org/2/library/platform.html#platform.system).

gvwilson:
I'm good either way --- only difference is that Linux and Mac both 
report 'posix' for os.name.  Your call...

wking:
On Thu, Jan 03, 2013 at 11:19:13AM -0800, Greg Wilson wrote:

> I'm good either way --- only difference is that Linux and Mac both 
> report 'posix' for os.name.  Your call...

Then if we want different instructions for Linux/Mac, we should
probably use `platform.system().lower()`.

wking:
I'm closing this PR. Future feedback should go in swcarpentry/boot-camps#12, which currently has a fairly complete implementation.

PR #93, opened 2012-12-15
Feedback on setup/index.html and setup/tests.html please?

caitsydney:
See my recent commit details for specific questions I have:
https://github.com/picken19/website/commit/6709732ef5fee15e67461382b0bcbe951bf427ea


jiffyclub:
I'm not a fan of the box. I'd suggest just trying to match styles with the rest of the site so that we keep things consistent. We'll probably keep iterating on the site design and do everything at once.

Are we really suggesting people use nano, vim, and emacs? These are the setup instructions I put together for my boot camps in October: http://swcarpentry.github.com/2012-10-caltech/#install

gvwilson:
> I'm not a fan of the box. I'd suggest just trying to match styles with
> the rest of the site so that we keep things consistent. We'll probably
> keep iterating on the site design and do everything at once.

I like the grid approach --- it's a lot easier for both instructors and 
learners to see what's there.  Matt, can you suggest a layout that would 
be equally easy to navigate?

> Are we really suggesting people use nano, vim, and emacs?

Let's get an easy-to-read format first (even I have lost track of what's 
in the existing setup instructions), and then we can vote and arm 
wrestle over specifics --- in particular, I think we need to decide what 
we're doing about Windows before we do anything else :-(

jiffyclub:
We could make some columns similar to what I did on this page: http://swcarpentry.github.com/2012-10-caltech/#install. Though three columns might get a bit tight.

We could also make a grid of links/buttons that either:
- link to an appropriate page
- pop open a dialog or tooltip with additional info

gvwilson:
Please no popups (dialogs and tooltips) - makes the site harder for the 
physically/visually disabled to work with.

caitsydney:
I can add the self-check scripts (#38, #39, #40) as an option in the tests.html page, if everyone is okay with that? I think then instructors could add their own scripts - I believe Fernando had some from his workshop - as they write them. The scripts seem like they may be comprehensive. Since we are using them, would we still need the specific testing instructions for how an attendee manually tests each thing they installed?

I've just been building the install info based on what was in the original page - I'll have a better idea of exactly what needs to go there after Wednesday, I'm hoping.

@gvwilson, I didn't know what to call the troubleshooting page. Will rename it. 

@jiffyclub, I really like the columns. I'll look at the code for it and see about doing something similar. I just wanted a way to divide the page more clearly for readability. A lot of info has to go on each page. I like your "bash," "git," etc. title boxes. That could work well.

Everyone - I'm going to need some sort of consensus about what people will need to install. Obviously people can use any text editor they prefer, but which one(s) should the website recommend? Same question for version control - do we need install instructions/use instructions for git/github, svn, BitBucket, and Mercurial? Something to talk about on Wednesday too, I think.

If we _do_ need all of those VC install/test instructions, would it make sense to have subpages specific to each type? The tests.html page is quickly getting unwieldy in length.  

setup
    >index.html
    >troubleshooting.html
    >install/
           >versioncontrol.html
           >texteditor.html
           >python.html
     >testing/    ---   has a toc linking to each type of software you will need
           >versioncontrol_test.html    -----   has a toc; attendees only focus on the vc they have been assigned by host
           >texteditor_test.html
           >python_test.html
           >scripts.html

jiffyclub:
The columns on that page are made using Bootstrap's grid system, docs here: http://twitter.github.com/bootstrap/scaffolding.html#fluidGridSystem

jiffyclub:
Some setup instructions will definitely be boot camp dependent. Not every boot camp has SQL, some of us do Mercurial, some of us do git...we were trying to come with a good way to make this very modular so instructors could put together just the instructions that applied to their boot camp.

PR #101, opened 2012-12-16
Add Twitter and RSS links to the footer

jiffyclub:


jiffyclub:
Here are some options.

Keep the label formatting but change the color:

Screen Shot 2012-12-15 at 4 05 59 PM

Use icons instead:

Screen Shot 2012-12-15 at 10 30 04 PM

gvwilson:
Please keep the label formatting with the change of color: I had to 
double-check what the 'g' icon was, and the 'mail' icon didn't make it 
clear that this was for mailing bug reports.
Thanks!

PR #115, opened 2012-12-17
index: Link "material" -> "contributing"

wking:
This raises the visibility on how you can get the source code, by restoring
the link which was removed in:

  commit ad60f32b7d8eac7868789de2b85bf402dca0c574
  Author: Greg Wilson gvwilson@third-bit.com
  Date:   Sat Dec 15 06:51:39 2012 -0500

```
Made the home page a little more interesting...
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -64,7 +64,8 @@
   <strong>Learn Online</strong><br />
   Our <a href="{{root_path}}/4_0/index.html">video lectures</a> cover
   all of our core topics and much else as well, and all of our
-  material can be freely re-used under
+  <a href="{{root_path}}/contributing.html#develop">material</a>
+  can be freely re-used under
   <a href="{{root_path}}/license.html">open access licenses</a>.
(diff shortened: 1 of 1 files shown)


gvwilson:
The text is about _using_ the material, but the link is to an explanation of how to _contribute_.  I think it should be left as-was.

wking:
On Mon, Dec 17, 2012 at 06:30:10AM -0800, Greg Wilson wrote:

> The text is about _using_ the material, but the link is to an
> explanation of how to _contribute_.  I think it should be left
> as-was.

Perhaps I should link into about/contributing.html#teach, which _is_
about using the material.  If you want a clean using/contributing
distinction, we could move the "Teach" blurb from contributing.html to
about/using.html or something…

wking:
Since the addition of `Join Us` in ad60f32b (#94), I'm happy enough with the visibility of the GitHub link, so I'm closing this.

PR #116, opened 2012-12-17
_base.html: add rel=vcs-* link to the Git repository

wking:
The rel=vcs-\* microformat [1](http://joeyh.name/rfc/rel-vcs/) makes it easy to find the versioned source of a
website without having to dig through human-readable descriptions.

### DIFF
--- a/_base.html
+++ b/_base.html
@@ -66,6 +66,9 @@
       <script src="http://html5shim.googlecode.com/svn/trunk/html5.js"></script>
     <![endif]-->
     {% endblock css %}
+		{% block rel_vcs %}
+		<link rel="vcs-git" href="https://github.com/swcarpentry/website.git" title="Software Carpentry Git repository" />
+		{% endblock rel_vcs %}
     {% block metadata %}{% endblock metadata %}
     {% block file_metadata %}{% endblock file_metadata %}
     {% block pagetitle %}<title>Software Carpentry: {{page.title}}</title>{% endblock pagetitle %}


bwinton:
I don't really follow Microformats much these days, but I will note that rel=vcs\* isn't listed on [their wiki page](https://en.wikipedia.org/wiki/Microformat), or in the list of specifications on [their home page](http://microformats.org/wiki/Main_Page#Specifications).

I would give this pull request a -1, at least until the rel=vcs-\* spec makes it through the microformat process.

Later,
Blake.

wking:
On Mon, Dec 17, 2012 at 06:51:27AM -0800, Blake Winton wrote:

> I don't really follow Microformats much these days, but I will note
> that rel=vcs\* isn't listed on [their wiki
> page](https://en.wikipedia.org/wiki/Microformat), or in the list of
> specifications on [their home
> page](http://microformats.org/wiki/Main_Page#Specifications).

I didn't realize that there were official microformats ;).  The
rel=vcs-\* spec is fairly old, stable, and easy to implement.  I don't
think that Joey is pushing it through any standardization process, but
I don't see why that should effect its usefulness.

bwinton:
It's probably because I'm unimaginative, but I don't see how it's useful.  Are there large sites which are providing this sort of information in either a standardized or unstandardized way?  Are there client programs which are trying to use it?  If it's so useful, why isn't anyone going through [the process](http://microformats.org/wiki/process) to standardize it?  (Reading the process page, it mainly seems to be a sanity check on whether a format should exist or not.  If this proposal hasn't gone through the process, then I have to ask questions like "Is this _actually_ solving a problem?  Which problem?  Haven't people solved it before, and why aren't we using their solution instead?"… ;)

wking:
On Mon, Dec 17, 2012 at 07:32:11AM -0800, Blake Winton wrote:

> It's probably because I'm unimaginative, but I don't see how it's
> useful.

When I first looked into SWC, I couldn't find the Subversion
repository.  I checked for a rel-vcs link, but there wasn't one.  The
GitHub repository is better advertised, but it's still nice to make
things machine readable.

> Are there large sites which are providing this sort of information
> in either a standardized or unstandardized way?

Not that I know of.

> Are there client programs which are trying to use it?

As mentioned on the rel-vcs page, joeyh@'s webcheckout uses it [1](http://git.kitenet.net/?p=mr.git;a=blob_plain;f=webcheckout;hb=HEAD).

> If it's so useful, why isn't anyone going through [the
> process](http://microformats.org/wiki/process) to standardize it?
> (Reading the process page, it mainly seems to be a sanity check on
> whether a format should exist or not.

I'm not sure what benefit you gain from standardizing the proposal
(getting support for this pull request? :p).

> If this proposal hasn't gone through the process, then I have to ask
> questions like "Is this _actually_ solving a problem?  Which
> problem?

Joey discusses the motivation in the blog post [2](http://joeyh.name/blog/entry/proposing_rel-vcs/) linked from the
rel-vcs page.

> Haven't people solved it before, and why aren't we using their
> solution instead?"… ;)

I'm not aware of other methods for specifying a website's source
repository in a machine readable way on the website itself.

gvwilson:
I agree with Blake: this one feels like a fringe technology.  If the only tool we know that uses it is the one written by its creator, and it's two years old, we'll leave it out for now.

PR #118, opened 2012-12-17
Python 3 compatibility for SWC website build tools

wking:
With these patches,

  $ make check

succeeds using Python 3.  Python 2.{6,7} should also work, but earlier
versions (e.g. Python 2.5) will choke on some newer syntax.

### DIFF
--- a/bin/chars.py
+++ b/bin/chars.py
@@ -13,8 +13,9 @@ for filename in sys.argv[1:]:
     tabs = True
     for (i, line) in enumerate(lines):
         if tabs and ('\t' in line):
-            print filename, i+1, 'tab'
+            print('{0} {1} {2}'.format(filename, i+1, 'tab'))
             tabs = False
         for (j, c) in enumerate(line):
             if ord(c) > 128:
-                print filename, i+1, j+1, ord(c), '&#%x;' % ord(c)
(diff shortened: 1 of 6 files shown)


gvwilson:
Do you have access to a Python 2.6.6 to test?  That's what we've got on the server right now.

wking:
On Mon, Dec 17, 2012 at 06:45:39AM -0800, Greg Wilson wrote:

> Do you have access to a Python 2.6.6 to test?  That's what we've got
> on the server right now.

I don't have access to 2.6.6, but I just tested it on 2.6.5.  It works
with 2.6.5, but I had to add three more patches to get 2.6.5
compatibility.  The patches (5f6154ac, a92101ec, and 1a82a291) are
syntax and ElementTree fixes for the old Python, and aren't related to
my Python 3 compatibility fixes.  Feel free to drop the new patches if
you're not interested in keeping compatibility with 2.6.5.

gvwilson:
Thanks very much for this --- ElementTree changed in 2.7, so if it works 
with 2.6.5, it'll work with 2.6.6. In theory.  I've put in a change req 
to Dreamhost for a Python upgrade, but I'm not holding my breath.

PR #122, opened 2012-12-18
RSS feed contains full HTML content for each RSS item.

pipitone:
I like to read an entire blog post in my feed reader, rather than having to open a new browser instance to do so.  At the moment, our RSS feed just delivers an excerpt of each post, so I'm forced to leave my reader.  

`<content:encoded>` tags are supposed to be the way to ship the full HTML content of an RSS item, according to this:
https://developer.mozilla.org/en-US/docs/RSS/Article/Why_RSS_Content_Module_is_Popular_-_Including_HTML_Contents

This commit introduces two subclasses of PyRSS2Gen to bin/compile.py to handle doing this correctly.  It also does some jinja2 mock-template rendering to get any variables in the blog post content to expand correctly.

### DIFF
--- a/bin/compile.py
+++ b/bin/compile.py
@@ -716,23 +716,46 @@ class PageFactory(object):
 
 #----------------------------------------
 
+class ContentEncodedRSSItem(RSSItem): 
+    def __init__(self, **kwargs): 
+        self.content = kwargs.get('content', None)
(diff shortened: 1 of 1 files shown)


jiffyclub:
Good idea!

PR #79, opened 2012-12-19
Move bootcamp setup instructions under 'bootcamp'

wking:
The top-level 'setup' directory should be moved under 'bootcamp', since that's what the instructions are for. This requires some changes to bin/compile.py (but it's an easy change).


wking:
On Fri, Dec 14, 2012 at 12:39:45PM -0800, Greg Wilson wrote:

> This requires some changes to bin/compile.py.

I didn't actually touch compile.py, but the move seems to work.  We
don't actually link into setup except from the individual boot camp
pages.

jiffyclub:
@picken19 is already at work modifying and updating the setup instructions. Maybe we should either let her take care of the move as part of her changes or wait until she's done.

gvwilson:
Agreed: let's assign to @picken19.
G

PR #133, opened 2012-12-20
Tidying up changes to compilation script.

gvwilson:
Minor formatting changes.

### DIFF
--- a/bin/compile.py
+++ b/bin/compile.py
@@ -172,9 +172,13 @@ class Application(object):
         """
         Parse command-line options.
         """
-        options, filenames = getopt.getopt(args, 'd:m:o:p:r:c:s:v')
+        options, filenames = getopt.getopt(args, 'c:d:m:o:p:r:s:v')
         for opt, arg in options:
-            if opt == '-d':
+            if opt == '-c':
+                assert self.icalendar_filename is None, \
+                       'iCalendar filename specified multiple times'
+                self.icalendar_filename = arg
+            elif opt == '-d':
                 self.today = arg
             elif opt == '-m':
                 assert self.metadata is None, \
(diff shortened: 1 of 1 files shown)


wking:
Docstring formatting like:

```
"""iCalendar generator for boot camps.

The format is defined in RFC 5545.
"""
```

follows PEP 257 [1](http://www.python.org/dev/peps/pep-0257/#multi-line-docstrings), while:

```
"""
iCalendar generator for boot camps.
The format is defined in RFC 5545: http://tools.ietf.org/html/rfc5545
"""
```

does not.  Not that it really matters unless your using a
docstring-processing tool like Sphinx, but it will be less surprising
for other contributors if we migrate to PEP 257 style docstrings at
some point.

gvwilson:
I've chosen consistency over PEP 257 for the moment; if you'd like to 
file an issue, we can fix the docstrings some time soon.

jiffyclub:
From PEP 257:

> The summary line may be on the same line as the opening quotes or on the next line.

And I personally much prefer the latter format:

```
"""
iCalendar generator for boot camps.
The format is defined in RFC 5545: http://tools.ietf.org/html/rfc5545
"""
```

wking:
On Thu, Dec 20, 2012 at 07:42:41AM -0800, Matt wrote:

> From PEP 257:
> 
> > The summary line may be on the same line as the opening quotes or on the next line.

I just learned something new :).

> And I personally much prefer the latter format:
> 
> ```
> """
> iCalendar generator for boot camps.
> The format is defined in RFC 5545: http://tools.ietf.org/html/rfc5545
> """
> ```

This is still not legal.  From PEP 257:

> The summary line may be used by automatic indexing tools; it is
> important that it fits on one line and is separated from the rest of
> the docstring by a blank line.

So perhaps we want (eventually):

```
"""
iCalendar generator for boot camps.

The format is defined in RFC 5545: http://tools.ietf.org/html/rfc5545
"""
```

?

jiffyclub:
Hey, I was just copying you. :wink: 

But yes, it is proper protocol to separate the summary line from the rest by an empty line.

PR #135, opened 2012-12-20
added a reference to the swcarpentry.github chicago requirements page.

katyhuff:
This is very minor and only affects the Uchicago bootcamp page. It adds a link to the setup page in the boot-camps repo, and fixes the date, which still said 19-20, but should be 12-13.

I'm not sure whether the swcarpentry ettiquette/protocol calls for this to be a pull request, but I tend to think even minor changes should be looked over by at least one other person, so I'll do it this way. Of course let me know if you guys think I should just push minor things like this in the future.


wking:
Looks good to me.

PR #142, opened 2012-12-21
Convert book Subversion chapter to Git

wking:
Now that we're using Git instead of Subversion, we should be teaching
Git instead of Subversion ;).

This initial pull request is a merger of my dvcs-content work with
some recent advances in the underlying Subversion chapter.  I don't
think it's ready for a merge yet, but I'm publishing this work in
progress so people can comment on the general layout while I polish
things up.

The major build-system change is the addition of my `swc-tags.py`
script for auto-generating command output.  This should make typos in
the examples much less likely.


gvwilson:
As discussed in point-to-point email, there are larger plans afoot (plus, we're going to continue offering Subversion for at least another year, possibly two, so we need to keep the existing material and build the Git stuff beside it).

wking:
This PR was closed, but I'm linking it to #1 anyway, in case the
eventual implementer wants to incorperate my existing work.

PR #159, opened 2012-12-29
Template variables not expanded in blog/index.html

wking:
bin/compile.py creates blog/index.html by stitching together the content of recent blog posts. However, the template variables in those fragments (such as {{root_path}}) are not being expanded, so links in blog/index.html are broken.


gvwilson:
Closed by 2b1e6dd.

PR #165, opened 2012-12-29
bootcamps: Add latlng metadata to several bootcamps

wking:
I don't have any personal connection to these boot camps, so I mostly used
the location of the hosting university.  Some of these boot camps actually
list addresses (e.g. 2013-01-CAMH) or link to pages that do
(e.g. 2013-02-amos), and I used that location to get more accurate 
coordinates where possible.

### DIFF
--- a/bootcamps/2013-01-CAMH.html
+++ b/bootcamps/2013-01-CAMH.html
@@ -4,6 +4,7 @@
   <meta name="date" content="January 2013" />
   <meta name="startdate" content="2013-01-24" />
   <meta name="enddate" content="2013-01-25" />
+  <meta name="latlng" content="43.6471184,-79.3942998" />
   <meta name="eventbrite_key" content="5086301264" />
 {% endblock file_metadata %}
 {% block content %}
--- a/bootcamps/2013-01-columbia.html
+++ b/bootcamps/2013-01-columbia.html
@@ -4,6 +4,7 @@
   <meta name="date" content="January 2013" />
   <meta name="startdate" content="2013-01-30" />
   <meta name="enddate" content="2013-02-02" />
+  <meta name="latlng" content="40.8080777,-73.9635678" />
 {% endblock file_metadata %}
 {% block content %}
 <p><strong>Where:</strong> Columbia University</p>
--- a/bootcamps/2013-01-tuebingen.html
+++ b/bootcamps/2013-01-tuebingen.html
@@ -4,6 +4,7 @@
   <meta name="date" content="January 2013" />
   <meta name="startdate" content="2013-01-25" />
   <meta name="enddate" content="2013-01-26" />
+  <meta name="latlng" content="48.5369804,9.058935" />
   <meta name="eventbrite_key" content="5014203618" />
 {% endblock file_metadata %}
 {% block content %}
--- a/bootcamps/2013-01-tum.html
+++ b/bootcamps/2013-01-tum.html
@@ -4,6 +4,7 @@
   <meta name="date" content="January 2013" />
   <meta name="startdate" content="2013-01-22" />
   <meta name="enddate" content="2013-01-23" />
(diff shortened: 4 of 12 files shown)


gvwilson:
Hm... can/should we put both "logical address" (human-readable) and 
"physical address" (lat/long) in headers?  It isn't always possible to 
go from one to the other without access to an email archive...

wking:
On Sat, Dec 29, 2012 at 10:34:04AM -0800, Greg Wilson wrote:

> Hm... can/should we put both "logical address" (human-readable) and 
> "physical address" (lat/long) in headers?  It isn't always possible to 
> go from one to the other without access to an email archive...

That's a good reason to list them both explicitly.  If you could go
back and forth automatically, it would be best to just use the human
readable metadata (venue).

Since the original author (mostly you?) left out the latlng info, I
assumed they weren't interested enough to care about fractions of
degrees.  If I'm wrong about that, I'm happy to re-roll with adjusted
latlngs, or to deprecate this pull request in favor of
organizer-authored pull requests.

gvwilson:
On 2012-12-29 1:51 PM, W. Trevor King wrote:

> On Sat, Dec 29, 2012 at 10:34:04AM -0800, Greg Wilson wrote:
> 
> > Hm... can/should we put both "logical address" (human-readable) and
> > "physical address" (lat/long) in headers? It isn't always possible to
> > go from one to the other without access to an email archive...
> 
> That's a good reason to list them both explicitly. If you could go
> back and forth automatically, it would be best to just use the human
> readable metadata (venue).

That's a bit "if" :-)  I've found in the past that several boot camp 
addresses (particularly outside the US) don't automatically resolve 
correctly in Google Maps.

> Since the original author (mostly you?) left out the latlng info, I
> assumed they weren't interested enough to care about fractions of
> degrees.

Wasn't a matter of "leaving out" --- it was more "didn't go back and do 
it when the feature was added".

> If I'm wrong about that, I'm happy to re-roll with adjusted
> latlngs, or to deprecate this pull request in favor of
> organizer-authored pull requests.

Not worth doing for past boot camps, but we can add an "address" field 
going forward?  Thoughts?

cheers,
Greg

wking:
On Sat, Dec 29, 2012 at 12:10:37PM -0800, Greg Wilson wrote:

> > If I'm wrong about that, I'm happy to re-roll with adjusted
> > latlngs, or to deprecate this pull request in favor of
> > organizer-authored pull requests.
> 
> Not worth doing for past boot camps, but we can add an "address" field 
> going forward?  Thoughts?

Ah, sorry.  I had thought you were arguing against both 'venue' and
'latlng', but you were actually arguing in favor of an additional
human-readable 'address' entry.  The distinction being that 'venue'
determines whether someone will attend (what they care about a month
in advance), and 'address' tells them exactly where to go (what they
care about 15 minutes in advance).  I think we should stick to one
location field (the current 'venue'), and more details should be free
form in the boot camp content.  Adding more structure would be
difficult (e.g. what if a boot camp is in different rooms on each
day?).

jiffyclub:
`venue` is used on the index page and `latlng` is used for the map. What purpose do you have in mind for an `address` field?

jiffyclub:
The ongoing discussion is sort of tangential to this PR, which looks good. I'm going to merge.

gvwilson:
On 2012-12-29 7:22 PM, Matt wrote:

> |venue| is used on the index page and |latlng| is used for the map. What
> purpose do you have in mind for an |address| field?

Intention is to use 'venue' as the name of the host institution, and 
'address' as the exact address for the class.  The former gets put in 
the bootcamps/index.html table; the latter would only show up on the 
page.  Assign this one to me...

gvwilson:
Let's stick to 'venue' (institution name, appears in 
bootcamps/index.html table) and 'latlng' (for map pins) for now: we need 
to get back to content creation...

PR #21, opened 2012-12-29
Write program to sync the Google Map and Google Calendar with actual workshop data

wking:
It would be _really_ nice if a single command could update the Google Map to show past and future bootcamps, and simultaneously update the Googel Calendar to show their dates --- doing this by hand is tedious and error-prone. This should also check information against EventBrite (particularly dates and venues).

### DIFF
--- a/_base.html
+++ b/_base.html
@@ -73,7 +73,7 @@
     {% block javascript %}{% endblock javascript %}
   </head>
 
-  <body>
+  <body {%- block onload %}{% endblock onload -%} >
     <div class="container shadow">
 
       <div class="banner">
--- a/bootcamps/index.html
+++ b/bootcamps/index.html
@@ -5,6 +5,78 @@
   <meta name="subglob" content="*-*-*.html" />
 {% endblock file_metadata %}
 
+{% block javascript %}
+  {{ super() }}
+  <script src="https://maps.googleapis.com/maps/api/js?v=3.exp&sensor=false"></script>
+  <script type="text/javascript">
+  function MarkerPin(color) {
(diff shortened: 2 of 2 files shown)


wking:
I've attached a pull request with an auto-generated Google Map.

The Google-Calendar portion of this issue was solved by PR #132 [1](https://github.com/swcarpentry/website/pull/132).

@jiffyclub, some of your work is in this commit.  Please review the
commit message and let me know if you think the attribution is
sufficient, or if you want me to split your work out into separate
commits.

Everyone else, please hold off merging until @jiffyclub chimes in.

jiffyclub:
Fine with me, thanks for getting this done.

jiffyclub:
I think the last thing to do here is to set up a Google calendar tracking http://software-carpentry.org/bootcamps.ics and embed it on http://software-carpentry.org/bootcamps/index.html. I could do that but I expect @gvwilson would prefer to be the admin.

That should fully automate the map and calendar.

gvwilson:
On 2012-12-29 8:46 PM, Matt wrote:

> I think the last thing to do here is to set up a Google calendar
> tracking http://software-carpentry.org/bootcamps.ics and embed it on
> http://software-carpentry.org/bootcamps/index.html. I could do that but
> I expect @gvwilson https://github.com/gvwilson would prefer to be the
> admin.
> 
> That should fully automate the map and calendar.

This connects with issue #22 (rationalizing online accounts): I think 
I'd like us to create a new Google account as 'admin', and have our map, 
our calendar, and our analytics all owned by that.

PR #169, opened 2012-12-30
Remove dates from calendar event summaries

wking:
The events in http://software-carpentry.org/bootcamps.ics have the date in their `SUMMARY` field, e.g. `SUMMARY:University of Oslo\, Sep 17-18\, 2012`. The summary field is used in the embedded Google calendar and the date adds a lot of text and makes it look crowded. This is on a calendar so it shouldn't be necessary to repeat the date in the summary.

Screen Shot 2012-12-29 at 11 12 49 PM


jiffyclub:
Thanks.

PR #168, opened 2012-12-30
Website build error in bootcamps/index.html

wking:
I'm getting this error trying to build the website with d916e72:

```
AssertionError: Unable to parse /Users/jiffyclub/projects/swc-website/build/bootcamps/index.html: not well-formed (invalid token): line 38, column 69
```

The code at that line is:

``` html
<script src="https://maps.googleapis.com/maps/api/js?v=3.exp&sensor=false"></script>
```

Anyone else have this problem and/or see what's causing this?

### DIFF
--- a/bootcamps/index.html
+++ b/bootcamps/index.html
@@ -9,6 +9,7 @@
   {{ super() }}
   <script src="https://maps.googleapis.com/maps/api/js?v=3.exp&amp;sensor=false"></script>
   <script type="text/javascript">
+  //<![CDATA[
   function MarkerPin(color) {
     var pin = new google.maps.MarkerImage(
         "http://chart.apis.google.com/chart?chst=d_map_pin_letter&chld=%E2%80%A2|" + color,
@@ -72,6 +73,7 @@
 {% endfor %}
 
   }
+  //]]>
   </script>
 {% endblock javascript %}
 
--- a/bootcamps/index.html
+++ b/bootcamps/index.html
@@ -7,7 +7,7 @@
 
 {% block javascript %}
   {{ super() }}
-  <script src="https://maps.googleapis.com/maps/api/js?v=3.exp&sensor=false"></script>
+  <script src="https://maps.googleapis.com/maps/api/js?v=3.exp&amp;sensor=false"></script>
   <script type="text/javascript">
   function MarkerPin(color) {
     var pin = new google.maps.MarkerImage(


gvwilson:
On 2012-12-29 11:06 PM, Matt wrote:

> I'm getting this error trying to build the website with d916e72
> https://github.com/swcarpentry/website/commit/d916e72:
> 
> |AssertionError: Unable to parse /Users/jiffyclub/projects/swc-website/build/bootcamps/index.html: not well-formed (invalid token): line 38, column 69
> |
> 
> The code at that line is:
> 
> <script src="https://maps.googleapis.com/maps/api/js?v=3.exp&sensor=false"></script>
> 
> Anyone else have this problem and/or see what's causing this?

I'm getting the same error. The problem is the ampersand character in 
the URL: change that to &amp; and the content parses just fine, but then 
of course it's the wrong bloody URL. _sigh_ I hate XML and character 
encoding and escaping rules so much...

Trevor?  Any idea how to build a URL that will get past Python's XML parser?

wking:
On Sun, Dec 30, 2012 at 05:35:09AM -0800, Greg Wilson wrote:

> I'm getting the same error. The problem is the ampersand character in 
> the URL: change that to &amp; and the content parses just fine, but then 
> of course it's the wrong bloody URL. _sigh_ I hate XML and character 
> encoding and escaping rules so much...

Ah, good catch.  Actually, it should be &amp; [1](http://www.w3.org/TR/xhtml1/#C_12).  I'm pushing a
patch on top of my original google-map branch.

gvwilson:
On 2012-12-30 8:50 AM, W. Trevor King wrote:

> On Sun, Dec 30, 2012 at 05:35:09AM -0800, Greg Wilson wrote:
> 
> > I'm getting the same error. The problem is the ampersand character in
> > the URL: change that to &amp; and the content parses just fine, but then
> > of course it's the wrong bloody URL. _sigh_ I hate XML and character
> > encoding and escaping rules so much...
> 
> Ah, good catch. Actually, it should be &amp; [1]. I'm pushing a
> patch on top of my original google-map branch.

When I changed the '&' to '&amp;' in build/bootcamps/index.html, the map 
wasn't populated with pins.

wking:
On Sun, Dec 30, 2012 at 05:51:42AM -0800, Greg Wilson wrote:

> On 2012-12-30 8:50 AM, W. Trevor King wrote:
> 
> > On Sun, Dec 30, 2012 at 05:35:09AM -0800, Greg Wilson wrote:
> > 
> > > I'm getting the same error. The problem is the ampersand character in
> > > the URL: change that to &amp; and the content parses just fine, but then
> > > of course it's the wrong bloody URL. _sigh_ I hate XML and character
> > > encoding and escaping rules so much...
> > 
> > Ah, good catch. Actually, it should be &amp; [1]. I'm pushing a
> > patch on top of my original google-map branch.
> 
> When I changed the '&' to '&amp;' in build/bootcamps/index.html, the map 
> wasn't populated with pins.

Probably because you're serving an XHTML webpage as HTML:

  $ wget -S -O /dev/null http://software-carpentry.org/bootcamps/index.html
  …
    HTTP/1.1 200 OK
    …
    Content-Type: text/html; charset=UTF-8
  …

If you want Firefox to interpret a local file as XHTML, you need to
give it the .xhtml extension.  Unfortunately it looks like '=' should
also be escaped:

  XML Parsing Error: not well-formed
  Location: file:///tmp/swc/bootcamps/index.xhtml
  Line Number 38, Column 70:
    <script src="https://maps.googleapis.com/maps/api/js?v=3.exp&sensor=false"></script>
  ---------------------------------------------------------------------^

I'm happy to patch this too, but perhaps we should bite the bullet and
switch to HTML 5.

wking:
On Sun, Dec 30, 2012 at 08:58:27AM -0500, W. Trevor King wrote:

> If you want Firefox to interpret a local file as XHTML, you need to
> give it the .xhtml extension.

And protecting against renames like this is a good reason to link to

  '.../bootcamps/'

instead of linking directly to

  '.../bootcamps/index.html'

wking:
On Sun, Dec 30, 2012 at 08:58:27AM -0500, W. Trevor King wrote:

> Unfortunately it looks like '=' should also be escaped:
> 
>   XML Parsing Error: not well-formed
>   Location: file:///tmp/swc/bootcamps/index.xhtml
>   Line Number 38, Column 70:
>     <script src="https://maps.googleapis.com/maps/api/js?v=3.exp&sensor=false"></script>
>   ---------------------------------------------------------------------^

Oops, I forgot to build the ampersand-escpated version.  This isn't
Firefox compaining about an unescaped `=`, it's complaining about the
invalid `&sensor=`.  From the XML 1.0 spec [1](http://www.w3.org/TR/2004/REC-xml-20040204/#NT-AttValue):

  [10]  AttValue   ::=   '"' ([^<&"] | Reference)\* '"'
                       | "'" ([^<&'] | Reference)\* "'"

so `=` is fine.

wking:
On Sun, Dec 30, 2012 at 08:58:27AM -0500, W. Trevor King wrote:

> On Sun, Dec 30, 2012 at 05:51:42AM -0800, Greg Wilson wrote:
> 
> > On 2012-12-30 8:50 AM, W. Trevor King wrote:
> > 
> > > On Sun, Dec 30, 2012 at 05:35:09AM -0800, Greg Wilson wrote:
> > > 
> > > > I'm getting the same error. The problem is the ampersand character in
> > > > the URL: change that to &amp; and the content parses just fine, but then
> > > > of course it's the wrong bloody URL. _sigh_ I hate XML and character
> > > > encoding and escaping rules so much...
> > > 
> > > Ah, good catch. Actually, it should be &amp; [1]. I'm pushing a
> > > patch on top of my original google-map branch.
> > 
> > When I changed the '&' to '&amp;' in build/bootcamps/index.html, the map 
> > wasn't populated with pins.
> 
> Probably because you're serving an XHTML webpage as HTML:

Ah, I've got something working with CDATA so I can use unescaped,
ampersands in the JavaScript.  I've pushed a rerolled PR which does
populate the map (even when served as text/html).

gvwilson:
> Ah, I've got something working with CDATA so I can use unescaped,
> ampersands in the JavaScript. I've pushed a rerolled PR which does
> populate the map (even when served as text/html).

I'm willing to accept CDATA as a temporary hack, but if HTML5 is the 
right long-term answer (and if Python's XML tools will actually accept 
it), how much other work needs to be done to get us there?

wking:
On Sun, Dec 30, 2012 at 06:57:45AM -0800, Greg Wilson wrote:

> > Ah, I've got something working with CDATA so I can use unescaped,
> > ampersands in the JavaScript. I've pushed a rerolled PR which does
> > populate the map (even when served as text/html).
> 
> I'm willing to accept CDATA as a temporary hack, but if HTML5 is the 
> right long-term answer (and if Python's XML tools will actually accept 
> it), how much other work needs to be done to get us there?

I have no idea ;).  I think the easiest way to check would be to
change the doctype in _base.html, build the site, push it somewhere
public, and start hitting http://validator.w3.org/.

Hopefully there won't be too much that's broken…

jiffyclub:
I think we should switch to HTML5 as soon as we can.

Where does the XML parsing come in, and could we use something like [Beautiful Soup](http://www.crummy.com/software/BeautifulSoup/) instead to make our lives easier?

wking:
On Sun, Dec 30, 2012 at 08:22:59AM -0800, Matt wrote:

> Where does the XML parsing come in, and could we use something like
> [Beautiful Soup](http://www.crummy.com/software/BeautifulSoup/)
> instead to make our lives easier?

If we're going to switch to an external dependency, I think
`lxml.html` [1](http://lxml.de/lxmlhtml.html) is a better choice than using Beautiful Soup directly.
If our input gets very soupy, it would be easy to switch to their
BeautifulSoup wrapper [2](http://lxml.de/elementsoup.html).

gvwilson:
On 2012-12-30 11:31 AM, W. Trevor King wrote:

> On Sun, Dec 30, 2012 at 08:22:59AM -0800, Matt wrote:
> 
> > Where does the XML parsing come in, and could we use something like
> > [Beautiful Soup](http://www.crummy.com/software/BeautifulSoup/)
> > instead to make our lives easier?
> 
> If we're going to switch to an external dependency, I think
> `lxml.html` [1] is a better choice than using Beautiful Soup directly.
> If our input gets very soupy, it would be easy to switch to their
> BeautifulSoup wrapper [2].
> 
> [1]: http://lxml.de/lxmlhtml.html
> [2]: http://lxml.de/elementsoup.html
1. We're parsing the generated pages to check that all internal links 
   resolve.  As a side effect, we're checking that they're all valid HTML.
2. I very much _don't_ want to use Beautiful Soup --- it's a great tool 
   for dealing with other people's HTML, but I want ours to be clean.
3. If I understand correctly, all I need to do to try out HTML5 is 
   change the top line in _base.html to:
   
    <!DOCTYPE html>

Trevor, anything else I should do?

Thanks,
Greg

wking:
On Sun, Dec 30, 2012 at 10:51:13AM -0800, Greg Wilson wrote:

> 1. We're parsing the generated pages to check that all internal links 
>    resolve.  As a side effect, we're checking that they're all valid HTML.
> 2. I very much _don't_ want to use Beautiful Soup --- it's a great tool 
>    for dealing with other people's HTML, but I want ours to be clean.
> 3. If I understand correctly, all I need to do to try out HTML5 is 
>    change the top line in _base.html to:
>    
>     <!DOCTYPE html>
> 
> Trevor, anything else I should do?

If you change the doctype so that you don't need to escape ampersands
in attribute values, you'll need to switch parsers.  If you're just
changing the doctype because HTML5 is cool, that's fine (and I agree
;), but it probably deserves a separate issue.

gvwilson:
On 2012-12-30 1:56 PM, W. Trevor King wrote:

> On Sun, Dec 30, 2012 at 10:51:13AM -0800, Greg Wilson wrote:
> 
> > 1. We're parsing the generated pages to check that all internal links
> >    resolve. As a side effect, we're checking that they're all valid HTML.
> > 2. I very much _don't_ want to use Beautiful Soup --- it's a great tool
> >    for dealing with other people's HTML, but I want ours to be clean.
> > 3. If I understand correctly, all I need to do to try out HTML5 is
> >    change the top line in _base.html to:
> > 
> > <!DOCTYPE html>
> > 
> > Trevor, anything else I should do?
> 
> If you change the doctype so that you don't need to escape ampersands
> in attribute values, you'll need to switch parsers.

To?  (If it means dropping ElementTree for something else, then it ain't 
gonna happen today :-)
- G

jiffyclub:
Why not [lxml.html](http://lxml.de/lxmlhtml.html) like @wking suggested? We're not trying to make our files valid XML, just valid HTML. Maybe we can take Trevor's fix now and then switch to something else later?

gvwilson:
On 2012-12-30 2:13 PM, Matt wrote:

> Why not lxml.html http://lxml.de/lxmlhtml.html like @wking
> https://github.com/wking suggested? We're not trying to make our files
> valid XML, just valid HTML. Maybe we can take Trevor's fix now and then
> switch to something else later?
1. What extra packages do would-be devs have to install, and how smooth 
   is that process?
2. What changes are required to our existing HTML pages to convert?
3. What changes are required to the scripts in 'bin'?
4. How different is this from what we teach in our HTML/XML material 
   (see book/web.html)?  I'd like to keep that in sync with the tools we 
   use ourselves in order to encourage people to transition from learners 
   to contributors.
   Thanks,
   Greg

wking:
On Sun, Dec 30, 2012 at 11:19:15AM -0800, Greg Wilson wrote:

> On 2012-12-30 2:13 PM, Matt wrote:
> 
> > Why not lxml.html...
> 1. What extra packages do would-be devs have to install, and how smooth 
>    is that process?

On Gentoo, `emerge -av lxml` ;).  For those with poor package managers:

  http://lxml.de/installation.html

> 1. What changes are required to our existing HTML pages to convert?

None that I know of, although HTML5 has some new tags for blog posts
(e.g. `article` [1](http://www.w3.org/TR/html5/sections.html#the-article-element)).  Still, all of that is optional.

> 1. What changes are required to the scripts in 'bin'?

lxml's interface is mostly an extension of the in-tree ElementTree
interface [2](http://lxml.de/api.html).  I imagine most changes here would be fairly trivial.

> 1. How different is this from what we teach in our HTML/XML material 
>    (see book/web.html)?

Not very.  The only serious difference is that lxml has a more relaxed
parser available that doesn't choke on some slightly-invalid XML.
Besides that, the basic API is very similar, and the mental model of
what's going on is identical.

gvwilson:
On 2012-12-30 2:38 PM, W. Trevor King wrote:

> On Sun, Dec 30, 2012 at 11:19:15AM -0800, Greg Wilson wrote:
> 
> > On 2012-12-30 2:13 PM, Matt wrote:
> > 
> > > Why not lxml.html...
> > 1. What extra packages do would-be devs have to install, and how smooth
> >    is that process?
> 
> On Gentoo, `emerge -av lxml` ;). For those with poor package managers:
> 
> http://lxml.de/installation.html

`easy_install lxml` seems to work on my Mac (OS X 10.8, Python 2.7.3).

> > 1. What changes are required to our existing HTML pages to convert?
> 
> None that I know of, although HTML5 has some new tags for blog posts
> (e.g. `article` [1]). Still, all of that is optional.

book/*.html already uses <section>, but yeah, defer.

> > 1. What changes are required to the scripts in 'bin'?
> 
> lxml's interface is mostly an extension of the in-tree ElementTree
> interface [2]. I imagine most changes here would be fairly trivial.

"Mostly" and "imagine" frighten me, lad :-)

> > 1. How different is this from what we teach in our HTML/XML material
> >    (see book/web.html)?
> 
> Not very. The only serious difference is that lxml has a more relaxed
> parser available that doesn't choke on some slightly-invalid XML.
> Besides that, the basic API is very similar, and the mental model of
> what's going on is identical.

OK, and just to confirm, this'll handle HTML5 out of the box, and we 
won't need to CDATA ampersands in URLs?

wking:
On Sun, Dec 30, 2012 at 11:42:12AM -0800, Greg Wilson wrote:

> On 2012-12-30 2:38 PM, W. Trevor King wrote:
> 
> > On Sun, Dec 30, 2012 at 11:19:15AM -0800, Greg Wilson wrote:
> > 
> > > 1. What changes are required to the scripts in 'bin'?
> > 
> > lxml's interface is mostly an extension of the in-tree ElementTree
> > interface [2]. I imagine most changes here would be fairly trivial.
> 
> "Mostly" and "imagine" frighten me, lad :-)

If the installation isn't a problem, then it should be easy to swap
parsers and check.  No use trying this though if there are other
barriers to lxml.

> OK, and just to confirm, this'll handle HTML5 out of the box, and we 
> won't need to CDATA ampersands in URLs?

  $ cat /tmp/test.html 
  <!DOCTYPE html>
  <html>
  <head>
          <link src="http://example.com/?wonky&url"/>
  </head>
  <body>
  </body>
  </html>
  $ cat test.py 
  #!/usr/bin/env python
  from lxml import etree
  parser = etree.HTMLParser()
  tree = etree.parse(open('test.html', 'r'), parser)
  print(etree.tostring(tree.getroot(),
                       pretty_print=True, method="html"))
  $ python3.2 test.py 
  b'<html>\n<head>\n<link src="http://example.com/?wonky&amp;url">\n<script type="text/javascript">\n\t\tvar a = "wierd < & string";\n\t</script>\n</head>\n<body>\n</body>\n</html>\n'
  $ python2.7 test.py 
  <html>
  <head>
  <link src="http://example.com/?wonky&amp;url">
  <script type="text/javascript">
                  var a = "wierd < & string";
          </script>
  </head>
  <body>
  </body>
  </html>

Note that <link> comes out without the built-in end tag (<link
... />), but it's still valid HTML5.  If you're just checking things
instead of editing them in the parser, this won't be a problem.

gvwilson:
On 2012-12-30 2:59 PM, W. Trevor King wrote:

> #!/usr/bin/env python
> from lxml import etree
> parser = etree.HTMLParser()
> tree = etree.parse(open('test.html', 'r'), parser)
> print(etree.tostring(tree.getroot(),
> pretty_print=True, method="html")) 
> OK, I did the following:
> 1. Install lxml using easy_install.
> 2. Strip down bin/util.py (see below).
> 3. Build the web site.
> 4. Run 'make book-gloss', 'make book-images', and a bunch of others.
> All seems to work. However, this parser _doesn't_ seem to complain about 
> ill-formed HTML --- if I put something like:

```
 <p>missing close tag
```

or:

```
 <div>whoops</span>
```

into an HTML file and parse it, it parses.  What's more, if I ask the 
lxml parser to read our Makefile, then write it out, what it does is 
wrap the whole Makefile in <html><body><p>...</p></body></html>. So, +1 
for handling the ampersand and not needing all the regexp crap to deal 
with special characters (I've tried a few, they all come through just 
fine), but -1 for not spotting badly-formed HTML.

Thoughts?

## G

from lxml import etree as ET

#-------------------------------------------------------------------------------

def read_xml(filename, mangle_entities=False):
     """
     Read in a document, returning the ElementTree doc node.
     """
     parser = ET.HTMLParser()
     with open(filename, 'r') as reader:
         return ET.parse(reader, parser)

#-------------------------------------------------------------------------------

def write_xml(filename, doc, unmangle_entities=False):
     """
     Write out a document.
     """
     with open(filename, 'w') as writer:
         writer.write(ET.tostring(doc.getroot()))

#-------------------------------------------------------------------------------

if **name** == '**main**':
     import sys
     if len(sys.argv) not in (2, 3):
         print >> sys.stderr, 'usage: util.py infile [outfile]'
     doc = read_xml(sys.argv[1])
     if len(sys.argv) == 3:
         write_xml(sys.argv[2], doc)

wking:
On Sun, Dec 30, 2012 at 05:23:13PM -0800, Greg Wilson wrote:

> So, +1 for handling the ampersand and not needing all the regexp
> crap to deal with special characters (I've tried a few, they all
> come through just fine), but -1 for not spotting badly-formed HTML.

If you want to write valid XML, then we escape ampersands, use CDATA,
etc., and a strict parser.  We can do this and use XHTML5 too, if we
want more modern tags.

If you want a parser that doesn't choke on invalid XML, you can use
lxml's HTMLParser, which will try fairly hard to not choke on invalid
input.

If you're looking for Goldilocks' “just right” level of strictness,
HTMLParser gives you some options [1](http://lxml.de/api/lxml.etree.HTMLParser-class.html).  `recover=False` may be what
you're looking for.  Otherwise I think your stuck with either extreme.

pipitone:
Looks like we're done here after the switch to LXML.  Closing.

Issue #8, opened 2012-12-08
Make sign-off for contributed content explicit


gvwilson:
Create some sort of sign-off for content contributors so that it's clear we have the right to use stuff.

wking:
Here's a dump of the earlier mailing list discussion into this issue:

```
On Sat, Dec 08, 2012 at 09:39:54AM -0800, Greg Wilson wrote:
> Create some sort of sign-off for content contributors so that it's
> clear we have the right to use stuff.

On Sat, Dec 08, 2012 at 11:54:44AM -0500, W. Trevor King wrote:
> On Sat, Nov 24, 2012 at 10:36:35AM -0500, W. Trevor King wrote:
> > On Mon, Oct 29, 2012 at 02:57:59PM -0400, Greg Wilson wrote:
> > > [reformatted to remove HTML]
> > > * Licensing
> > >   * Existing content is CC-BY / MIT licensed
> > >   * Do *not* want to have to manage content with multiple licenses
> > >   * Need sign-off from contributors
> > >   * Linux Submitting Patches [1] guide and GitHub's Contributing
> > >     Guidelines [2] are both examples we can copy.
> > >
> > > [1]: http://kernel.org/doc/Documentation/SubmittingPatches
> > > [2]: https://github.com/blog/1184-contributing-guidelines
> > 
> > I've been trying to figure out the licensing for Linux's Developer
> > Certificate of Origin (listed in SubmittingPatches).  As best I can
> > determine, it is CC BY-SA 2.5 generic, copyright by the OSDL [1].
> > 
> > I've asked for clarification on this issue [2], because if it is true
> > and we want to stick to pure CC BY 3.0 unported, the Linux DCO is
> > *not* and example we can copy.
> > 
> > [1]: http://web.archive.org/web/20070306195036/http://osdlab.org/newsroom/press_releases/2004/2004_05_24_dco.html
> > [2]: Thread:  http://thread.gmane.org/gmane.linux.kernel/1397613
> >      Message: http://article.gmane.org/gmane.linux.kernel/1400065
> 
> I emailed some of the related Linux/Git devs to get explicit
> confirmation of the licensing for the DCO/s-o-b.  The clearest
> response I got was:
> 
> On Sat, Nov 24, 2012 at 02:39:39PM -0800, Greg Kroah-Hartman wrote:
> > All you should care about is the DCO itself, if you want to use it
> > somewhere else.
> >
[...]

PR #171, opened 2012-12-30
Updating embedded calendar to follow http://software-carpentry.org/bootcamps.ics

jiffyclub:
I made a Google calendar on the admin account that follows our generated calendar file and switched the embedded calendar to point to that. I also added some margin space to the bottom of the map div to put some room between the map and calendar.

### DIFF
--- a/_base.html
+++ b/_base.html
@@ -131,13 +131,13 @@
           <img src="{{root_path}}/img/creative-commons-attribution-license.png" alt="Creative Commons Attribution License" />
         </a>
         <span class="pull-right">
-          <a class="label"
+          <a class="label swc-blue-bg"
              href="https://twitter.com/swcarpentry">Twitter</a>
-          <a class="label"
+          <a class="label swc-blue-bg"
              href="https://github.com/swcarpentry">GitHub</a>
-          <a class="label"
+          <a class="label swc-blue-bg"
              href="http://software-carpentry.org/feed.xml">RSS</a>
-          <a class="bugreport label"
+          <a class="bugreport label swc-blue-bg"
              href="mailto:{{contact_email}}?subject=bug%20in%20{{filename}}">
              Bug Report
           </a>
--- a/bootcamps/index.html
+++ b/bootcamps/index.html
@@ -119,15 +119,15 @@ attendees to have an enjoyable and fulfilling experience.</p>
 {% endfor %}
 </table>
 
-<div id="map_canvas" style="width: 98%; height: 500px; margin-left:auto; margin-right: auto"></div>
+<div id="map_canvas" style="width: 98%; height: 500px; margin-left:auto; margin-right: auto; margin-bottom: 25px;"></div>
 
 <p>
-  <iframe style="border: 0pt none;"
-          src="https://www.google.com/calendar/embed?src=workshops%40software-carpentry.org&amp;ctz=America/Toronto"
-          frameborder="0"
-          scrolling="no"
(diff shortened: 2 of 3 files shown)


wking:
The patch looks good to me, but I haven't built and tested it.

gvwilson:
If it builds and works for @jiffyclub, please merge it.
- G

gvwilson:
Builds and works for me; only thought is that the calendar title is the URL of the .ics file as plain text (see screenshot): can it either be removed (context makes calendar purpose clear) or changed to be a hyperlink wrapped around the URL?

Screen Shot 2012-12-30 at 7 48 40 PM

jiffyclub:
I just figured out it is possible to set the title to something else (though not sure about a hyperlink), but I just took it off since it probably is superfluous. I also took off the print icon and little pull down button on the far right so it looks like this:

Screen Shot 2012-12-30 at 8 03 19 PM

wking:
You might want to hold off on the s-o-bs until we close #8 [1](https://github.com/swcarpentry/website/issues/8), it's
looking like we may not want them at all.

jiffyclub:
Could do something like this:

Screen Shot 2012-12-30 at 9 28 54 PM

jiffyclub:
Done.

wking:
Looking at 65b6dcfb75, I think we should stick to `label` instead of
adding `swc-label`.  This is why CSS is “cascading”, right?

jiffyclub:
I don't want to actually change the style of `label`, though. That could have other uses. It might be more appropriate to rename `swc-label` to `swc-background` since all it's modifying is the background color.

wking:
On Tue, Jan 01, 2013 at 07:37:22AM -0800, Matt wrote:

> I don't want to actually change the style of `label`, though. That
> could have other uses.

Such as what?  If you want some labels to have a dark background
color, then you probably want all the labels to have the dark
background color.  If you only want some labels to have a dark
background color (e.g. in the footer), you can mask by div ID (as was
done before 65b6dcfb).

If you want to make things “SWC Blue” without forcing a semantic
interpretation, I'd be fine with a `swc-blue` class, but claiming that
some labels are “SWC labels” while others are not smells funny.

jiffyclub:
There are a whole suite of label styles built into Bootstrap, I'm not going to override them: http://twitter.github.com/bootstrap/components.html#labels-badges

I will change the class name to something more informative, though.

wking:
On Tue, Jan 01, 2013 at 08:14:10AM -0800, Matt wrote:

> There are a whole suite of label styles built into Bootstrap, I'm
> not going to override them:
> http://twitter.github.com/bootstrap/components.html#labels-badges

Ah, thanks for the link.  Do we use these labels anywhere?  I don't
see why the default label shouldn't have a blue background.

jiffyclub:
No, I don't think they're in use anywhere else on the site. I just don't want to change defaults and confuse people later. If anyone wants to make a matching label they should be able to copy the existing examples.

wking:
On Tue, Jan 01, 2013 at 08:43:02AM -0800, Matt wrote:

> No, I don't think they're in use anywhere else on the site. I just
> don't want to change defaults and confuse people later. If anyone
> wants to make a matching label they should be able to copy the
> existing examples.

I expect future devs are more likely to not care what color label they
have, and defaulting to SWC branding seems like a good idea.  However,
I don't care enough to argue it further ;).

jiffyclub:
Okay, I amended the last commit so there's now an `swc-blue-bg` class instead of `swc-label`, and updated the current label instances to have that class.

jiffyclub:
I think this is settled enough to merge, I'll hit the button.

Issue #184, opened 2013-01-03
Firefox add-on instruction should be "Web Developer" not "Tools"


rmdickson:
http://software-carpentry.org/setup/common.html says "When Firefox restarts, Select Tools => SQLite Manager", but this is unavailable in Firefox 17.0.1.  

Should read "When Firefox restarts,select Web Developer => SQLite Manager"

ethanwhite:
This is actually a little bit OS dependent because Ubuntu by default doesn't use the big orange button in order to allow integration with it's HUD system, in which case it is located under Tools, but I definitely agree that this should be updated to the more common form perhaps with a parenthetical for Ubuntu users.

wking:
I think this was one of the side effects of the whole “version numbers
are scary” mindset [1](https://bugzilla.mozilla.org/show_bug.cgi?id=678775) ;).

wking:
Ethan's commit fixes this issue, but we forgot to close it until now.

PR #186, opened 2013-01-03
Updating instructions for running the Firefox SQLite Manager

ethanwhite:
As described in #184 the instructions for running SQLite Manager were out of date.

### DIFF
--- a/setup/common.html
+++ b/setup/common.html
@@ -16,6 +16,6 @@
 <li>Double-click to select <strong>SQLite Manager 0.7.7</strong></li>
 <li>Click <strong>Install</strong></li>
 <li>Click <strong>restart now</strong></li>
-<li>When Firefox restarts, Select <strong>Tools</strong> =&gt; <strong>SQLite Manager</strong></li>
+<li>When Firefox restarts, Select either <strong>Web Developer</strong> =&gt; <strong>SQLite Manager</strong> or <strong>Tools</strong> =&gt; <strong>SQLite Manager</strong></li>
 </ul>
 {% endblock content %}


wking:
I'd remove “on Ubuntu” and just list the alternative location.  This
is probably just a difference between Firefox versions.  I certainly
can't think of a reason why only Ubuntu would patch Firefox to change
this location.

ethanwhite:
It's not a version difference (I'm running 17 on both Windows 7 and Ubuntu), but it does turn out to be possible to have the alternative form on other operating systems, it's just that the default is one way on Ubuntu (related to the integration with Unity's universal menu bar and HUD) and the other way on everything else I've seen. So, yes, we should change it.

ethanwhite:
Done. Is the a preference for amending vs. rebasing for this sort of simple change?

wking:
On Thu, Jan 03, 2013 at 12:00:58PM -0800, Ethan White wrote:

> Is the a preference for amending vs. rebasing for this sort of
> simple change?

What's the difference?  I think the Ubuntu reference should be
squashed out (which it looks like you did).  You can do that with an
amend or with a second commit followed by a squash rebase.  The result
is the same (but the amend is easier to do in this case).

On an unrelated note (possibly for another PR?), do we want to use
`=&gt;` here instead of the more semantic (but possibly less well
supported?) `→` (a.k.a '&#x2192;', Unicode's RIGHTWARDS ARROW).

ethanwhite:
OK, so they're the same. That answers my question. Thanks.

The `=&gt;` things seems like something separate, on which I have no opinion.

gvwilson:
I'm in favor of &rarr; instead of =&gt;
- G

ethanwhite:
I'm happy to make the change here if that's appropriate or we can do it in a separate PR. Just let me know.

wking:
On Thu, Jan 03, 2013 at 12:41:35PM -0800, Ethan White wrote:

> I'm happy to make the change here if that's appropriate or we can do
> it in a separate PR. Just let me know.

I'll open a separate PR.  `git grep` turns up quite about 36 of these
;).

PR #187, opened 2013-01-03
Convert `=&gt;` to the more semantic `&rarr;`.

wking:
`&rarr;` is defined in the HTML 4 spec [1](http://www.w3.org/TR/html4/sgml/entities.html):

  <!ENTITY rarr CDATA "&#8594;" -- rightwards arrow, U+2192 ISOnum -->

This should probably wait until after an lxml migration (see
discussion in #168) so we don't confuse Python's stdlib xml parser.

### DIFF
--- a/setup/common.html
+++ b/setup/common.html
@@ -10,12 +10,12 @@
 <p>The SQLite Manager add-on can be installed through the "Add-ons Manager" in Firefox.</p>
 <ul>
 <li>Start Firefox</li>
-<li>Select <strong>Tools</strong> =&gt;<strong> Add-ons</strong></li>
+<li>Select <strong>Tools</strong> &rarr;<strong> Add-ons</strong></li>
 <li>In the search field, enter <code>sqlite</code></li>
 <li>Click the search icon</li>
 <li>Double-click to select <strong>SQLite Manager 0.7.7</strong></li>
 <li>Click <strong>Install</strong></li>
 <li>Click <strong>restart now</strong></li>
-<li>When Firefox restarts, Select <strong>Tools</strong> =&gt; <strong>SQLite Manager</strong></li>
+<li>When Firefox restarts, Select <strong>Tools</strong> &rarr; <strong>SQLite Manager</strong></li>
 </ul>
 {% endblock content %}
--- a/setup/index.html
+++ b/setup/index.html
@@ -262,9 +262,9 @@ $ firefox
 <h2>Firefox SQLite manager for a relational database</h2>
 <p>Once Firefox is running:</p>
 <ul>
-<li>Select <strong>Tools</strong> =&gt; <strong>SQLite Manager</strong></li>
+<li>Select <strong>Tools</strong> &rarr; <strong>SQLite Manager</strong></li>
 <li>SQLite Manager window appears</li>
-<li>Select <strong>Database</strong> =&gt; <strong>New database</strong></li>
+<li>Select <strong>Database</strong> &rarr; <strong>New database</strong></li>
 </ul>
 <p><a href="http://code.google.com/p/sqlite-manager/">http://code.google.com/p/sqlite-manager/</a></p>
 {% endblock content %}
--- a/setup/osx.html
+++ b/setup/osx.html
@@ -27,7 +27,7 @@
 <ul>
 <li>Click <strong>Applications</strong></li>
 <li>Click <strong>XCode</strong></li>
-<li>Select <strong>XCode</strong> =&gt; <strong>Preferences...</strong></li>
+<li>Select <strong>XCode</strong> &rarr; <strong>Preferences...</strong></li>
 <li>Click <strong>Downloads</strong></li>
 <li>Select <strong>Command Line Tools</strong></li>
 <li>Click <strong>Install</strong></li>
--- a/setup/terminal.html
+++ b/setup/terminal.html
@@ -8,11 +8,11 @@
 <p>To check your software or install it for a boot camp requires you to use a  terminal window (also commonly called a command-line <a href="http://en.wikipedia.org/wiki/Shell_(computing)">shell </a>or <a href="http://en.wikipedia.org/wiki/Terminal_emulator">terminal emulator</a>).</p>
 <p>Here are ways to get a terminal window for popular operating systems.</p>
 <p><strong>To get a terminal window in Ubuntu:</strong></p>
-<p>Select the menu <strong>Applications</strong> =&gt; <strong>Accessories</strong> and select <strong>Terminal.</strong> You should see a terminal window like,</p>
+<p>Select the menu <strong>Applications</strong> &rarr; <strong>Accessories</strong> and select <strong>Terminal.</strong> You should see a terminal window like,</p>
 <p><a href="{{root_path}}/files/2012/10/ubuntu-terminal.jpg"><img title="ubuntu-terminal" src="{{root_path}}/files/2012/10/ubuntu-terminal-300x197.jpg" alt="Ubuntu terminal" width="300" height="197" /></a></p>
 <p><strong>To get a terminal window in RedHat/Scientific Linux 6:</strong></p>
 <p>This depends on what windowing system you are using whether it be KDE, GNOME or X-windows.</p>
-<p>For example, for GNOME select the menu <strong>Applications</strong> =&gt; <strong>System Tools</strong> and select <strong>Terminal</strong>. You should see a terminal window like,</p>
+<p>For example, for GNOME select the menu <strong>Applications</strong> &rarr; <strong>System Tools</strong> and select <strong>Terminal</strong>. You should see a terminal window like,</p>
 <p><img src="{{root_path}}/files/2012/10/gnome-terminal-300x195.jpg" alt="Gnome terminal" /></p>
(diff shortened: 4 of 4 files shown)


blog/2011/08/introducing-programming-a-different-way.html
```diff
@@ -416,8 +416,8 @@
     sum = 0
     for i in range(limit):
         sum += data[i]
-    print data, "=&gt;", sum
```
gvwilson, line 419: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

blog/2011/08/introducing-programming-a-different-way.html
```diff
@@ -416,8 +416,8 @@
...
-<span style="color: darkred; background-color: lightyellow;">[] =&gt; 0 [10] =&gt; 10 [10, 20] =&gt; 30 [10, 20, 30] =&gt; 60 [10, 20, 30, 40] =&gt; 60</span></pre>
+    print data, "&rarr;", sum
+<span style="color: darkred; background-color: lightyellow;">[] &rarr; 0 [10] &rarr; 10 [10, 20] &rarr; 30 [10, 20, 30] &rarr; 60 [10, 20, 30, 40] &rarr; 60</span></pre>
```
gvwilson: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

blog/2011/08/introducing-programming-a-different-way.html
```diff
@@ -457,8 +457,8 @@
         for i in range(i, limit):
             sum += data[i]
         result.append(sum)
-    print data, "=&gt;", result
```
gvwilson, line 460: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

blog/2011/08/introducing-programming-a-different-way.html
```diff
@@ -480,7 +480,7 @@
 <p>How we actually understand this program is:</p>
 <pre>for data in test_cases:
     result = sum_by_threes(data)
-    print data, "=&gt;", result
```
gvwilson, line 483: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

book/funclib.html
```diff
@@ -913,7 +913,7 @@
 print average3(1, 5)
 <span class="err">Traceback (most recent call last):
   File "src/funclib/average-3-wrong.py", line 4, in &lt;module&gt;
-    print 1, 5, '=&gt;', average3(1, 5)
```
gvwilson, line 916: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

book/funclib.html
```diff
@@ -973,10 +973,10 @@
     return result / len(values)
 
 for test in [[1.0], [1.0, 2.0], [1.0, 2.0, 5.0]]:
-    print test, '=&gt;', average_list(test)
```
gvwilson, line 976: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

book/funclib.html
```diff
@@ -973,10 +973,10 @@
...
-<span class="out">[1.0] =&gt; 1.0
```
gvwilson, line 977: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

book/funclib.html
```diff
@@ -1064,12 +1064,12 @@
 
 <pre src="src/funclib/average_with_defaults.py">
 numbers = [1.0, 2.0, 5.0]
-print '(', numbers, ') =&gt;', average_list(numbers)
```
gvwilson, line 1067: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

book/funclib.html
```diff
@@ -1296,12 +1296,12 @@
       </p>
 
 <pre src="src/funclib/sign.py">
-print -5, '=&gt;', sign(-5)
```
gvwilson, line 1299: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

book/funclib.html
```diff
@@ -1463,12 +1463,12 @@
         return 0
 <span class="comment">#    return 1</span>
 
-print -5, '=&gt;', sign(-5)
```
gvwilson, line 1466: This is a snippet of Python code, so we need to display '=>' (because that's what the Python actually contains).

gvwilson:
- Where '=>' appears as a string in a Python program to be printed, it should be left alone.
- Where it appears in setup/*.html, the conversion is good.

wking:
On Sun, Jan 06, 2013 at 05:21:02AM -0800, Greg Wilson wrote:

> - Where '=>' appears as a string in a Python program to be printed,
>   it should be left alone.

Oops.  Sloppy me not checking all my automatic replacements.  How do
you feel about replacing the Python occurences with `\u2192'`?

wking:
On Sun, Jan 06, 2013 at 08:36:34AM -0500, W. Trevor King wrote:

> On Sun, Jan 06, 2013 at 05:21:02AM -0800, Greg Wilson wrote:
> 
> > - Where '=>' appears as a string in a Python program to be printed,
> >   it should be left alone.
> 
> Oops.  Sloppy me not checking all my automatic replacements.  How do
> you feel about replacing the Python occurences with `\u2192'`?

On Sun, Jan 06, 2013 at 05:18:14AM -0800, Greg Wilson wrote:

> > @@ -416,8 +416,8 @@
> >      sum = 0
> >      for i in range(limit):
> >          sum += data[i]
> > -    print data, "=&gt;", sum
> 
> This is a snippet of Python code, so we need to display '=>'
> (because that's what the Python actually contains).

Oops, I think I gave in too quickly ;).  This is Python code embedded
in `<pre>` tags, so it _should_ contain `&rarr;`.  Because of Python
2.x's wonky unicode handling, maybe the line should read:

```
print data, u"&rarr;", sum
```

wking:
On Sun, Jan 06, 2013 at 05:47:00AM -0800, Greg Wilson wrote:

> -1: I don't want us to have to explain Unicode escape sequences that 
> early in the show.  If we could actually trust all the editors people 
> use, and legacy tools like Cygwin's 'cat', to display non-7-bit 
> characters correctly, I'd just put the ' →' directly in the source, but 
> hey, I'd like a unicorn too...

Python 3.x and `# -*- coding: utf-8 -*-` make life so much nicer ;),
but I'm ok rolling this back.  Rebased PR pushed.

wking:
```
On Sun, Jan 06, 2013 at 06:14:14AM -0800, Greg Wilson wrote:
> On 2013-01-06 8:50 AM, W. Trevor King wrote:
> > Oops, I think I gave in too quickly ;). This is Python code embedded
> > in `<pre>` tags, so it *should* contain `&rarr;`. Because of Python
> > 2.x's wonky unicode handling, maybe the line should read:
> >
> > print data, u"&rarr;", sum
>
> Nope --- when people compare the text they see in the browser page to 
> the .py file on disk, they should see exactly the same thing. 
> Discrepancies are puzzling, and we really don't want them more puzzled 
> than they have to be.
```

If we'd cone this way, it would have been easy enough to update the
external script.  Maintaining this sort of thing is why I wrote my
`swc-tags.py` script in #142 [1](https://github.com/wking/swc-website/blob/book-git/bin/swc-tags.py).  The real problem is this:

On Sun, Jan 06, 2013 at 06:15:55AM -0800, Greg Wilson wrote:

> On 2013-01-06 9:04 AM, W. Trevor King wrote:
> 
> > On Sun, Jan 06, 2013 at 05:47:00AM -0800, Greg Wilson wrote:
> > 
> > > -1: I don't want us to have to explain Unicode escape sequences that
> > > early in the show. If we could actually trust all the editors people
> > > use, and legacy tools like Cygwin's 'cat', to display non-7-bit
> > > characters correctly, I'd just put the ' →' directly in the source, but
> > > hey, I'd like a unicorn too...
> > 
> > Python 3.x and `# -*- coding: utf-8 -*-` make life so much nicer ;),
> > but I'm ok rolling this back. Rebased PR pushed.
> 
> On the one hand, yes, for people who understand what that's all about. 
> On the other hand, try drawing the concept map of what people have to 
> understand in order to understand that magic line --- it's not small.  
> For the same reason, we don't use '#! /usr/bin/env python' and 'chmod 
> +x' to make scripts directly runnable: they're both the right thing to 
> do, but there's such a thing as being prematurely right :-)

I understand there's only so much you can do in a day ;).  Still,
Python 3 make this easy, since UTF-8 is the default [3](http://www.python.org/dev/peps/pep-3120/).  Then for
people on sane systems that use UTF-8 and Python 3, there's nothing
you _need_ to discuss.

The problem is people on oddball systems that don't use UTF-8 ;).

Out of curiosity, does Cygwin's `cat` really care about file encoding?
I would expect problems with the terminal displaying unicode to be
more likely.

> p.s. this feels like a good topic for a blog post --- may I quote our 
> conversation?

Quote away ;).  The `coding` comment is defined in PEP 263 [2](http://www.python.org/dev/peps/pep-0263/) and
UTF-8 became the default in Python 3 after PEP 3120 [3](http://www.python.org/dev/peps/pep-3120/), if you want
to link people to details.

Issue #161, opened 2012-12-23
Create blog index by author and topic


gvwilson:
We currently index by date alone; need to have an index by author, and another by topic.

wking:
I'm still not entirely convinced that rolling our own blogging engine
is the way to go here.

wking:
On Thu, Jan 03, 2013 at 08:03:40AM -0500, Greg Wilson wrote:

> …did you ever look at Pelican (a Python compile-your-blog tool)?
> Jon P. and I explored using it to build our website instead of
> rolling our own, but it looked like the hard part --- extracting and
> collating metadata --- wasn't handled at all, so building the
> extensions we'd need would be just as much work as what we're doing
> now.

It looks like arbitrary metadata is pretty easy to use [1](http://docs.getpelican.com/en/3.1.1/faq.html#can-i-use-arbitrary-meta-data-in-my-templates).  Also, the
engine building the `blog/` directory does not need to be the same as
the engine building the rest of the site.  One problem with the
current `compile.py` is that one a page declaring a `subfile` or
`subglob` gets easy access to that (group of) page(s).  This leads to
hackery like the existing `create_rss` and `ICalendarWriter` in
`compile.py`, which each use their own `Instances` hack to get access
to a particular group of pages.  `blog/index.html` is already mostly
calendar indexing (because splitting that index out into
`blog/calendar-index.html` or some such would require a new
`compile.py` hack).  Adding more index styles would require either
more `compile.py` hacks, or a fairly thorough `compile.py`
restructuring.

I'll give Pelican another look and see how easy it is to stretch over
our needs…

wking:
On Thu, Jan 03, 2013 at 01:22:59PM -0800, Greg Wilson wrote:

> On 2013-01-03 8:25 AM, W. Trevor King wrote:
> 
> > On Thu, Jan 03, 2013 at 08:03:40AM -0500, Greg Wilson wrote:
> > 
> > > …did you ever look at Pelican (a Python compile-your-blog tool)?
> > > Jon P. and I explored using it to build our website instead of
> > > rolling our own, but it looked like the hard part --- extracting and
> > > collating metadata --- wasn't handled at all, so building the
> > > extensions we'd need would be just as much work as what we're doing
> > > now.
> > 
> > It looks like arbitrary metadata is pretty easy to use [1].
> 
> But only within the same doc --- as soon as I want doc X to use metadata 
> from doc Y (e.g., to use information about particular bootcamps in 
[...]

PR #192, opened 2013-01-06
Adding key to Google Map (closes #61).

### LINKED ISSUE #61: Add key to Google Map explaining pin color
We're using blue and purple to show future and past workshops; the map should have a key.

gvwilson:
- Adding table below Google Map displaying icons uses and their meanings.
- Added two icon images to img/icons for display in key.


wking:
On Sun, Jan 06, 2013 at 05:01:16AM -0800, Greg Wilson wrote:

> - Added two icon images to img/icons for display in key.

If you're going to store local copies of these icons for use in the
key, you might as well use the local copies when you generate the map
(we currently request images from chart.apis.google.com, see
`MarkerPin()` in `bootcamps/index.html`.

wking:
On Sun, Jan 06, 2013 at 05:43:25AM -0800, Greg Wilson wrote:

> On 2013-01-06 8:31 AM, W. Trevor King wrote:
> 
> > On Sun, Jan 06, 2013 at 05:01:16AM -0800, Greg Wilson wrote:
> > 
> > > - Added two icon images to img/icons for display in key.
> > 
> > If you're going to store local copies of these icons for use in the
> > key, you might as well use the local copies when you generate the map
> > (we currently request images from chart.apis.google.com, see
> > `MarkerPin()` in `bootcamps/index.html`.
> 
> Good point: please make the fix and push it?  (I don't know how to 
> insert a marker pin generated by Javascript into the DOM.)

It's not a particularly important fix, just for internal consistency.
This is less of a priority for me than getting the conversion to
Pelican in place before the trunk gets too far out ahead of me ;).  I
can look into it after #161 goes in.

I also think you should document the source of the icons (if you did
download them from Google) and check the terms of service, since I
didn't check their license when linking to them originally.

PR #197, opened 2013-01-09
Ability to turn off "Past boot camps" pins in map

jiffyclub:
In the map on http://software-carpentry.org/bootcamps/index.html it would be nice if users could turn off the "Previous Boot Camps" pins. It's nice to see where we've been but it's not necessarily interesting to a potential participant who is looking for a boot camp to attend (assuming that is the audience for this page).

### DIFF
--- a/bootcamps/index.html
+++ b/bootcamps/index.html
@@ -25,6 +25,14 @@
     })
   }
 
+  function toggle_marker_visibility(marker) {
+    if (marker.getVisible()) {
+      marker.setVisible(false);
+    } else {
+      marker.setVisible(true);
+    };
+  }
+
   function initialize_map() {
     var mapOptions = {
       zoom: 2,
@@ -46,6 +54,8 @@
     var map = new google.maps.Map(document.getElementById('map_canvas'),
         mapOptions);
 
+    var past_markers = [];
+
 {% for bootcamp in page.children %}
   {% if bootcamp.latlng %}
     var marker = new google.maps.Marker({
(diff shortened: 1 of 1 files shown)


wking:
On Tue, Jan 08, 2013 at 10:52:13AM -0800, Amy Brown wrote:

> In the map on http://software-carpentry.org/bootcamps/index.html it
> would be nice if users could turn off the "Previous Boot Camps"
> pins. It's nice to see where we've been but it's not necessarily
> interesting to a potential participant who is looking for a boot
> camp to attend (assuming that is the audience for this page).

That's a good idea, but I don't know enough about the Google Maps
JavaScript API to know how to code it up.

jiffyclub:
Since the site is static this is a tough thing to do. In a quick glance through the Maps API docs I can't find a way to do it at all. A couple options:
- don't show past bootcamps at all
- limit the past bootcamps to something like 10 events
- make the past bootcamps icon smaller, maybe a dot instead of a marker?

wking:
On Tue, Jan 08, 2013 at 01:36:04PM -0800, Matt Davis wrote:

> Since the site is static this is a tough thing to do. In a quick
> glance through the Maps API docs I can't find a way to do it at all.

It seems that removing markers is possible:

  http://stackoverflow.com/questions/1544739/google-maps-api-v3-how-to-remove-all-markers

You'd need to tie a javascript function to a button on the page to do
this though, and we'd need to keep a list of past-boot-camp markers.

jiffyclub:
I might see a way to do this, actually:
- make an array of `pastMarkers`
- make a function that toggles their `setVisible` method
- [add a button](https://developers.google.com/maps/documentation/javascript/controls#CustomControls) to the map that calls the toggle function

I'll see if I can give this a try tonight.

jiffyclub:
The map looks like so, now:

Screen Shot 2013-01-08 at 9 35 07 PM

I made it so that past boot camps **do not** show up by default, you have to click the button.

I found out it's possible to set the `z-index` of the markers so that some show up above others, which would allow us to make it so that future boot camps always appear in front of markers for past boot camps, but I figure with the toggle switch that's not necessary.

Tested in recent versions of Chrome, Firefox, and Safari.

gvwilson:
I like it --- please push to swcarpentry/master and we'll deploy.
Thanks!
Greg

PR #203, opened 2013-01-09
Yeah, I know, I should split this...

gvwilson:
...but my attempts to do so have borked, and we need to get this into the repo ASAP.


wking:
On Wed, Jan 09, 2013 at 08:04:23AM -0800, Greg Wilson wrote:

> ...but my attempts to do so have borked, and we need to get this
> into the repo ASAP.

If you have a dirty tree from work on X and you see an unrelated
problem Y that could use a quick fix, the recipe is:

  $ git stash save
  $ git checkout -b fix-Y origin/master
  $ sensible-editor file-with-Y
  …fix Y…
  $ git commit -am 'Fix Y'
  $ git checkout fix-X
  $ git stash pop

If Y is related to X, but not close enough go into the same commit,
there's:

  $ sensible-editor file-with-Y
  …fix Y…
  $ git add -p file-with-Y
  …select hunk with the Y fix…
  $ git commit -m 'Fix Y'

If Y is already glommed onto an X-fixing commit, see the `SPLITTING
COMMITS` section of the `git rebase --help`.

Issue #59, opened 2012-12-12
Add drop-down menus to web site


gvwilson:
(split from #55)

Everyone but @gvwilson wants drop-down menus on the web site.

jiffyclub:
I'm not so sure about that "everyone". And for several of the menu items I'm not actually sure what would go in the drop down. You've done a good job putting links to the most useful information at the top of each sub-page so I'm not sure drop-downs would add much navigational help.

karthik:
Since the framework is Bootstrap, this is easy and [readily available](http://twitter.github.com/bootstrap/components.html#dropdowns). I'm indifferent on this matter but it would make site navigation easier. `About`, for example, could have a dropdown with 
- Credits
- Our Team
- Our Audience
- Testimonials
- How to Contribute
- Reading 

With bootcamps it could be **upcoming** and **past**. I'd say this isn't a priority but might be useful to implement down the line.

gvwilson:
On 2012-12-12 1:31 PM, Karthik Ram wrote:

> |About|, for example, could have a dropdown with
> - Credits
> - Our Team
> - Our Audience
> - Testimonials
> - How to Contribute
> - Reading

In that particular case, I'd actually get rid of the 'About' page 
entirely (since the key information is on the home page).

karthik:
Definitely, that makes sense.

caitsydney:
Agreed about the ease of site navigation. 

For example, getting to Boot Camp > FAQ and Bootcamp > How to Run a Boot Camp can only be accomplished by first going to the Boot Camp page and then looking through the text for the appropriate links. Might increase site traffic to these sub-pages if drop-down menus are implemented? Not sure if there's a reason that the menus haven't been implemented this way already.

jiffyclub:
I think @gvwilson had some technical aversion to them based on his experience on the existing site. But with Bootstrap I don't worry too much about that, I think their code is high quality.

We just need a good map of what goes in the drop downs. The existing menus on http://software-carpentry.org are rather overloaded and I don't want to just duplicate those.

karthik:
Agreed. The current version is daunting. All the lectures don't need to be on there. And License could be moved to the footer (a location where people typically look for it) to save space up top.

LJWilliams:
Maybe not dropdown boxes, but a navigation bar underneath the main header and links to the relevant subpages?

PR #205, opened 2013-01-09
Adding drop downs to Boot Camp, Blog, and About menu items

jiffyclub:
Dropdown menu items are:
- Boot Camps
  - About
  - Future Boot Camps
  - Map
  - Calendar
  - Past Boot Camps
- Blog
  - Recent Posts
  - All Posts
- About
  - FAQ
  - Sponsors
  - Team
  - Contributing

### DIFF
--- a/_base.html
+++ b/_base.html
@@ -98,6 +98,18 @@
         <div class="nav-collapse collapse">
           <ul class="nav">
             <li id="nav-home"><a href="{{root_path}}/index.html">Home</a></li>
+            <li id="nav-about" class="dropdown">
+              <a class="dropdown-toggle" data-toggle="dropdown" href="#">
+                About
+                <b class="caret"></b>
+              </a>
+              <ul class="dropdown-menu" role="menu">
+                <li><a href="{{root_path}}/about/faq.html">FAQ</a></li>
+                <li><a href="{{root_path}}/about/credits.html">Sponsors</a></li>
+                <li><a href="{{root_path}}/about/team.html">Team</a></li>
+                <li><a href="{{root_path}}/about/contributing.html">Contributing</a></li>
+              </ul>
+            </li>
             <li id="nav-lessons"><a href="{{root_path}}/4_0/index.html">Lessons</a></li>
             <li id="nav-bootcamps" class="dropdown">
               <a class="dropdown-toggle" data-toggle="dropdown" href="#">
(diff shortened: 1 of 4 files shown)


jiffyclub:
This was originally brought up eons ago in #59.

gvwilson:
Nice --- a few thoughts:
1. Click on 'About' to bring up the menu, then click again without moving the mouse: 'About' stays highlighted, even thought we're still on the home page.
2. Given this, 'About' should be right after 'Home'.
3. Yeah, we need to add a menu for 'Lessons'.  It's OK that this isn't dynamically generated --- the 4_0 stuff isn't going to change.  But the first item in the list needs to be 'Index' (or something like that) to go to 4_0/index.html.
Thanks!

jiffyclub:
1. That's the hover. All the menu items highlight when the mouse is over them.
2. Can do.
3. You mean a menu like this?
   - Version Control
   - Python
   - Shell
   - Testing
   - ...
   
   That seems a little overkill given the list at the top of 4_0/index.html.

jiffyclub:
I'm not sure there's anything that can be done about that, it's the clicks that change the open/closed state. Clicking anywhere will collapse the menu.

jiffyclub:
We'd probably have to dig into the Bootstrap JavaScript to change that. I'm not actually writing JS here, just adding classes to things in the menubar. Bootstrap takes care of the rest.

jiffyclub:
I really think the current behaviour is fine (and works exactly like the menubar on my Mac).

jiffyclub:
Yes, but if I move the mouse off the menubar the menus stay open.

jiffyclub:
Cool. Trying to modify the default behaviour of the JS libraries is not something I'm keen to dive into.

gvwilson:
On 2013-01-11 10:16 AM, Matt Davis wrote:

> Cool. Trying to modify the default behaviour of the JS libraries is 
> not something I'm keen to dive into.
> 
> Hm: I googled for 'jquery pulldown menu', and the first example I found 
> does this:

http://www.queness.com/resources/html/dropdownmenu/index.html#

??

jiffyclub:
And Bootstrap dropdowns do this: http://twitter.github.com/bootstrap/javascript.html#dropdowns

I don't think we're going to change this without writing JS, and I'd really rather not. (If someone else wants to...)

jiffyclub:
Cool. If someone really wants to practice their JS and DOM skills in the future we can point them at this.

PR #224, opened 2013-01-17
Move license links

jiffyclub:
I moved the "License" link from the navbar under "About", removed the CC-BY image from the lower-left, and added a "License" badge among the others in the lower-right.

### DIFF
--- a/_base.html
+++ b/_base.html
@@ -105,6 +105,7 @@
               </a>
               <ul class="dropdown-menu" role="menu">
                 <li><a href="{{root_path}}/about/faq.html">FAQ</a></li>
+                <li><a href="{{root_path}}/license.html">License</a></li>
                 <li><a href="{{root_path}}/about/credits.html">Sponsors</a></li>
                 <li><a href="{{root_path}}/about/team.html">Team</a></li>
                 <li><a href="{{root_path}}/about/contributing.html">Contributing</a></li>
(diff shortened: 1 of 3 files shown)


wking:
On Wed, Jan 16, 2013 at 05:14:11PM -0800, Matt Davis wrote:

> - Moved "License" navbar link under "About".

Tabs indenting the new `</li>` line should be replaced with spaces.

jiffyclub:
Arrgh, that file was full of tabs. Think they're all gone now.

wking:
On Thu, Jan 17, 2013 at 06:12:06AM -0800, Matt Davis wrote:

> Arrgh, that file was full of tabs. Think they're all gone now.

I only saw that one… Anyhow it looks good to me now, although I'd
squash 5e89c921068fea6247be8848d989d7c8ffe95042 into
1f3bd278a7df513268f50680fb3688c378bb1f88 before merging.

jiffyclub:
That's because I'd already removed the rest. ;-)

wking:
On Thu, Jan 17, 2013 at 06:55:40AM -0800, Matt Davis wrote:

> That's because I'd already removed the rest. ;-)

Ah, it looks like a lot of that happened in
7ff8c45d19b5435bf6d2ca390a867310894fb5e8.  Maybe we need some
pre-commit hooks to check for these SWC-style issues ;).

PR #228, opened 2013-01-18
Adding schedule to Paris boot camp

gvwilson:


wking:
On Fri, Jan 18, 2013 at 03:24:33AM -0800, Greg Wilson wrote:

> - Schedule for Paris bootcamp

It's possible you left some dots out of the “Structuring, documenting,
…” line.  And you should also probably be using `&hellip;` instead of
`...`.

PR #250, opened 2013-01-25
blog/2013/01/visualizing: Link to "Categorization of Used Nuclear..."

wking:
If we reference a paper, we might as well link to it.  Hopefully the ornl.gov
link will be pretty stable.


gvwilson:
Thanks -- please go ahead and merge.
G

PR #257, opened 2013-01-28
Blog post about novelty, efficiency, and trust.

gvwilson:


wking:
On Mon, Jan 28, 2013 at 06:05:09AM -0800, Greg Wilson wrote:

> - Blog post about novelty, efficiency, and trust

Looks reasonable to me.

PR #256, opened 2013-01-28
Drop down menu on SWC website not behaving as expected on mobile browsers

jiffyclub:
There seems to be a bug in the behaviour of the drop-down nav menu for the Software Carpentry website under Safari on an iPhone running iOS 6.0.1.

Clicking on the icon renders the menu top-level labels as expected, and touching the label (e.g. Boot Camps) expands out the next level labels properly.

However, if you try and press one of these labels (e.g. Upcoming Boot Camps), instead of loading that page, it acts as if you had pressed the top-level label that would have been in that area of the screen previously (in this case Badges).

**This means you can't get to the page you want to using the drop down menu, and therefore can't get to any page not linked in the body without resorting to a search engine or direct URL, on an iPhone**

It appears to afflict all browsers based on Apple's default iOS webkit implementation on an iPhone (I haven't tried on an iPad or Android device). The bug appears to be consistent across pages and menu items on the site.


npch:
This might be considered a part of [Issue 52](/swcarpentry/website/issues/52), but could do with being temporarily patched if possible.

ethanwhite:
This bug also effects my Android phone.

npch:
I've edited the issue title to reflect the more universal nature of the bug.

jiffyclub:
I'll try updating to Bootstrap 2.2.2, which according to twitter/bootstrap#4550 looks like it should contain a fix.

jiffyclub:
Updating to Bootstrap 2.2.2 seems to have done the trick, you can test it at http://dev.software-carpentry.org. I've tested it in Chrome on Android 4, don't have an iOS device handy to try it on.

jiffyclub:
On Android's built-in browser there's still an odd issue where you have to press on the sub-menu for half a second to get it to remain open, otherwise it opens and immediately closes. This seems related to twitter/bootstrap#6488, which appears slated to go into Bootstrap 2.3.0 sometime in the indefinite future. If this is a serious problem on iOS it should be possible to replicate the diff in twitter/bootstrap#6488 as a temporary workaround.

jiffyclub:
I've tested this on iOS 6 and it seems to work fine. The site has some other, cosmetic issues, though...

gvwilson:
Should the merge wait until the "cosmetic issues" are resolved, or go ahead now?

jiffyclub:
Go ahead and merge, that's work for another time.

npch:
Confirm that this merge has corrected the original issue for me.

PR #258, opened 2013-01-30
Blog post about boot camp at Mozilla

gvwilson:


blog/2013/01/boot-camp-at-mozilla.html
```diff
@@ -0,0 +1,56 @@
...
+<meta name="title" content="A Boot Camp at Mozilla" />
+<meta name="post_date" content="2013-01-30" />
+<meta name="category" content="boot-camp" />
+<meta name="category" content="venues/mozilla" />
+{% endblock file_metadata %}
+{% block content %}
+
+<p>We ran a two-day boot camp at Mozilla's office in Toronto last week for people from local research hospitals.  It seemed to go well: only 28 of the 37 who'd registered showed up, but with a couple of exceptions because of scheduling conflicts, everyone who showed up on day 1 came back on day 2.  Many thanks to Dhavide Aruliah, Yele Bonilla, Mike Conley, Gabriel Devenyi, Fan Dong, and Blake Winton for helping out.</p>
+
+<table>
```
jiffyclub: If you add `class="table"` here you'll get a pretty Bootstrap table instead of the user-agent default.

blog/2013/01/boot-camp-at-mozilla.html
```diff
@@ -0,0 +1,56 @@
...
+  <tr>
```
jiffyclub: If you wrap this in `<thead>` (and add the table class) you can drop the `<strong>` tags, CSS will take care of it.

blog/2013/01/boot-camp-at-mozilla.html
```diff
@@ -0,0 +1,56 @@
...
+    <td><strong>Good</strong></td>
+    <td><strong>Bad</strong></td>
+  </tr>
+  <tr>
+    <td valign="top">
```
jiffyclub: The `valign` attribute is obsolete in HTML5, and unnecessary anyway. The Bootstrap `table` class specifies the vertical alignment as part of the CSS.

blog/2013/01/boot-camp-at-mozilla.html
gvwilson: I put <thead>...</thead> around the first row, and <tbody>...</tbody> around the other row --- gives me a horizontal line between the two, but the 'Good' and 'Bad' in the first row are _not_ bolded.

jiffyclub: My bad, you also want `<th>` elements instead of `<td>`.

jiffyclub:
I've made all my suggested changes in gvwilson/website#1.

jiffyclub:
The diff looks good. Too bad 62edf11 made it into the history, but I don't think you'll get rid of that without rebasing or something.

jiffyclub:
Actually, just doing `git rebase fff2e89` seems to pretty things up.

PR #267, opened 2013-02-01
CONTRIBUTING.md: Change 'clone' to 'fork' for GitHub duping

wking:
This is the lingo that GitHub uses, and for me the instructions read more
clearly with a local clone and a remote fork.


pipitone:
Right on!  Thanks Trevor.

PR #270, opened 2013-02-01
Added check-bare to the Makefile for doing an unvalidated build

pipitone:
The reason I'm adding this option is because whilst adding features to the website, I often go through iterations of adding content and rebuild to check that the website looks okay. I find myself always stopping make when it gets to the validation step because early on I just don't care if it's valid or not (it probably is, and if not, I'll fix it up later once the site looks as it should).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,10 +103,14 @@ install :
 ## check        : rebuild entire site locally for checking purposes.
 check : $(STATIC_DST) $(OUT_DIR)/.htaccess
 	@make ascii-chars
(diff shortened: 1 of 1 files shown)


jiffyclub:
Oh thank god, running the build script takes way too damn long.

pipitone:
Another way to speed things up might be to get the makefile to work under parallelisation (i.e. with the -J flag).  Make can do this when a target has multiple dependencies, and we could also use gnu parallel to do some of the work for targets that process bunches of files (I'm looking at you `make check-links`)

PR #298, opened 2013-02-12
Adding Libarkin report to files/papers

amyrhoda:
I put a copy of the Libarkin report in files/papers.


about/faq.html
```diff
@@ -120,11 +120,13 @@
...
 
-<p>Yes: two independent studies in 2012 found that people really were
-learning useful things, and
-<a href="{{root_path}}/about/testimonials.html">these testimonials</a>
-from past participants show just how much difference a little training
-can make.</p>
+<p>Yes: two independent studies in 2012 (<a
+	href="{{root_path}}/files/papers/aranda-assessment-2012-07.pdf">Aranda</a> and
+<a
+	href="{{root_path}}/files/papers/libarkin-assessment-report-2012-06.pdf">STEM
```
gvwilson, line 126: "STEM E.D." ?

amyrhoda, line 126: I looked for a reference in the file itself and there isn't one. The 
head of the document looks like the headers from an email message, and 
the From: line says STEM E.D.  Julie Libarkin's name doesn't appear in 
the document anywhere.

gvwilson, line 126: Hm --- @ctb and @picken19, how should we cite this?

PR #308, opened 2013-02-18
Added pages for upcoming UK boot camps

mikej888:
EGI Forum, Manchester 11 April (boot camp "highlights" day)
Oxford DTCs, Oxford, 19-20 May
ICSS, Southampton, June (dates to be decided)
Wellcome Trust, University of Exeter, November (dates to be decided depending on interest)
Wellcome Trust, University of Exeter (Tremough Campus), November (dates to be decided depending on interest)

### DIFF
--- a/bootcamps/2013-04-egi-forum.html
+++ b/bootcamps/2013-04-egi-forum.html
@@ -1,0 +1,38 @@
+{% extends "_bootcamp.html" %}
+{% block file_metadata %}
+  <meta name="venue" content="EGI Forum, Manchester" />
+  <meta name="date" content="April 2013" />
+  <meta name="startdate" content="2013-04-11" />
+  <meta name="latlng" content="53.467102, -2.233958" />
+{% endblock file_metadata %}
+{% block content %}
+<p><strong>Where:</strong> EGI Forum, Manchester.</p>
+<p><strong>When:</strong> April 11th, 2013 between 10:00 and 17:30</p>
+<p>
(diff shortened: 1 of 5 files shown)


pipitone:
@mikej888 Do you want to use use eventbrite to manage registration for these bootcamps?  There is a software-carpentry admin account we could arrange this through.

gvwilson:
I prefer to keep managing things through EventBrite unless there are 
powerful reasons to do otherwise (some sites insist on using their 
universities' reg. systems, for ex.).
- G

pipitone:
@gvwilson Makes sense. 

So, perhaps my question to @mikej888 is then: would you like to include links to an eventbrite sign up page on each of these bootcamp pages?

mikej888:
EGI Forum doesn't need one - it'll be open to whoever turns up on the day.
Oxford DTCs and Soton are internals - I've to ask whether they want EventBrite (as we're doing for the Manchester internal).
Wellcome Trust I'll set up EventBrite in due course.
I just wanted to get the pages in place afore updating and extending them as needed,

mikej888:
For Wellcome Trust, they initially want to assess how much interest there is in their local groups and from this they'll then narrow down a date. If it would be more suitable to just have them e-mail me and I just keep a list in SVN, instead of using EventBrite, then please let me know.

mikej888:
After discussion with @amyrbrown, we won't use EventBrite until dates are confirmed.

PR #327, opened 2013-02-28
New page for University College London April camp

amyrhoda:
This bootcamp is split; it's not over two consecutive days. It's not obvious how to change the H1 header to read "University College London: Apr 04 and 08, 2013" so I didn't -- is it worth the bother?


gvwilson:
What we really need is an "anddate" metadata tag, but since we're not doing any more work on our current website compiler (bring on Pelican!), yes, this is fine.

PR #330, opened 2013-03-01
Adding a page that displays upcoming boot camps' staffing

gvwilson:
Our database of who's teaching what, when, is stored in a separate (private) 'admin' repo, since it includes things like personal email addresses.  There's a script in that repo that generates a static HTML page 'bootcamps/staffing.html', which we will check into this repo periodically.

### DIFF
--- a/bootcamps/staffing.html
+++ b/bootcamps/staffing.html
@@ -1,0 +1,126 @@
+{% extends "_base.html" %}
+{% block file_metadata %}
+  <meta name="title" content="Boot Camp Staffing" />
+{% endblock file_metadata %}
+{% block content %}
+<h3>Last Updated 2013-03-01</h3>
+<table cellpadding="5">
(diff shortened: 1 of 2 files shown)


wking:
On Fri, Mar 01, 2013 at 12:39:09PM -0800, Greg Wilson wrote:

> Our database of who's teaching what, when, is stored in a separate
> (private) 'admin' repo, since it includes things like personal email
> addresses.  There's a script in that repo that generates a static
> HTML page 'bootcamps/staffing.html', which we will check into this
> repo periodically.

Can you commit the script and an SQL dump of this database with some
dummy data (in assets?)?  It would be nice to limit the secret keeping
to the sensitive data (and have a public script, schema, etc.).

jiffyclub:
Do you also want to add a link to this from somewhere?

jiffyclub:
What's its purpose if people can't browse to it?

jiffyclub:
The Bootcamps menu drop down and/or somewhere on `bootcamps/index.html`.

Issue #328, opened 2013-03-01
Need a subscribe/unsubscribe page for announcements email list


amyrhoda:
See http://wiki.dreamhost.com/KB_/_Web_Programming_/_CGI%2C_PHP_%26_Databases_/_Announcement_Mailing_List#How_can_people_subscribe_to_my_mailing_list.3F

wking:
Closed with 8218f76398bcb3f9bc2bb32a9a110395c1e12bb8 merging #331.

PR #331, opened 2013-03-01
Added mailing list info/sign up to contributing page, and some clean up.

pipitone:
Specifically, I merged the Report Bugs and Suggest Improvements into the Create
and new Discuss section. I removed stuff about creating new slides and audio
because that's no longer relevant.

Additionally, I added a Discuss section with links to the blog, announcement
mailing list, and the discussion mailing list. There is a form on the page
now that lets people (un)subscribe from the announcement mailing list.

### DIFF
--- a/about/contributing.html
+++ b/about/contributing.html
@@ -8,58 +8,70 @@
 
 {% block content %}
 
-<p>Software Carpentry is an open project. Like all such projects, it needs your help to grow, and you can contribute in many ways, large and small.</p>
+<p>Software Carpentry is an open project. Like all such projects, it needs your
+help to grow, and you can contribute in many ways, large and small.</p>
 
 <dl>
 
-  <dt id="bugs">Report Bugs</dt>
-  <dd>
-    If you see a mistake in our lectures or example programs, please add a comment to the appropriate page, or <a href="mailto:{{contact_email}}">mail us</a>, and we'll fix it as quickly as we can.
-  </dd>
-
-  <dt id="improve">Suggest Improvements</dt>
-  <dd>
-    Is one of our lessons too long?  Do they move too quickly or too slowly, or give too much or too little detail?  Did we miss anything important, or include things you're sure you'll never need?  If so, please <a href="mailto:{{contact_email}}">let us know</a>.  When doing so, please tell us a bit about your background and needs so that we know how best to interpret your feedback.
-  </dd>
-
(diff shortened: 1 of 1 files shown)


pipitone:
I should note that this PR fixes #328.

wking:
It looks like you're missing some <p> tags around the final paragraph
(“If you see a mistake…”).  I'm too fried to go through the rest of
this in detail now though :p.  I'll take a more thorough look
tomorrow.

pipitone:
@wiking Nice catch, thanks. Fixed.

wking:
Looking at this with `--word-diff=color` makes it a lot easier to
review ;).  Comments:
- You add spaces around an `&mdash;`, but Wikipedia says that it's
  conventional to use em dashes without spaces [1](http://en.wikipedia.org/wiki/Dash).
- I'd probalby say “we'd be glad to provide” instead of “we'd be glad
  to help provide”.
- You probably don't need the space in “requests. </p>”, not that it's
  doing any harm…  There's another gratuitous space later:
  “decisions. </p>”.
- Removing the slides/audio request may be a serious enough change
  that it needs a separate PR (or OK from Greg :p).  We seem to be
  moving away from this format, but I'm not sure yet where we're
  moving to ;).
- You leave out a number of URL fields (url, unsuburl, …) that are
  used in the example template [2](http://wiki.dreamhost.com/KB_/_Web_Programming_/_CGI,_PHP_&_Databases_/_Announcement_Mailing_List#How_can_people_subscribe_to_my_mailing_list.3F).  This is fine with me, but more
  folks with more branding instincts might want to draw up
  SWC-specific versions of the result pages.

pipitone:
Thanks @wking. I've fixed the typos you pointed out.  

I should have mentioned that I did talk with Greg about removing the slides/audio requests and got his okay. Perhaps a separate discussion on what we, as a group, are moving towards is in order.

wking:
On Sat, Mar 02, 2013 at 08:49:48AM -0800, Jon Pipitone wrote:

> I should have mentioned that I did talk with Greg about removing the
> slides/audio requests and got his okay.

In that case, I think this is good to merge.  I'll do it in a few
hours if no other issues come up, but feel free to jump in before me
;).

> Perhaps a separate discussion on what we, as a group, are moving
> towards is in order.

Yeah, this PR is probably not the place for that discussion ;).

PR #333, opened 2013-03-03
Blog post on teaching with ipythonblocks at UW

jiffyclub:
Text of post:
## 

The recent [University of Washington boot camp](http://software-carpentry.org/blog/2013/03/washington-went-well.html) was my first chance to
try teaching Python with [ipythonblocks](http://penandpants.com/2013/01/10/ipythonblocks-a-visual-tool-for-practicing-python/), and it was a great success!

ipythonblocks demo image.

The students reported that the immediate feedback from ipythonblocks was a
great way to see what their code was doing. We were able to move quickly
through topics like for-loops and if-statements in relative comfort, and even
cover slicing with ease.

I had always hoped that the opportunities for creativity with ipythonblocks
would appeal to students. At the UW boot camp we did observe people,
especially those who grasped the concepts quickly, striking off on their own
to experiment with larger grids and more complex algorithms for varying block
color. I think this is a great advantage of the visual nature of ipythonblocks;
there is only such much creative fun to be had playing with lists, but far
more possibilities with what is essentially a grid of pixels.

My only regret is that we ran out of time so the students didn't get to make
[Starry Night](http://nbviewer.ipython.org/urls/raw.github.com/jiffyclub/ipythonblocks/master/demos/starry_night_to_text.ipynb) or practice with files and functions as much as I'd have liked.
I think it wouldn't be too hard to expand the lesson I'd planned to spend all
day playing with [ipythonblocks](https://github.com/jiffyclub/ipythonblocks).


wking:
On Sat, Mar 02, 2013 at 05:30:54PM -0800, Matt Davis wrote:

> The students reported that the immediate feedback from ipythonblocks
> was a great way to see what their code was doing.

Yay!

> I think it wouldn&#39;t be too hard to expand…

Is there a reason to use `&#39;` instead of `'`?

jiffyclub:
`&#39;` gets rendered as an apostrophe instead of a single-quote.

wking:
On Sat, Mar 02, 2013 at 06:21:48PM -0800, Matt Davis wrote:

> `&#39;` gets rendered as an apostrophe instead of a single-quote.

That's odd.  `'` _is_ an apostrophe (U+0027 APOSTROPHE) :p.  Is there
some not-overly-smart smart-quote conversion going on?  If we're
dropping into escaped characters, the Unicode notes for APOSTROPHE
suggest:

  U+2019 RIGHT SINGLE QUOTATION MARK is preferred for apostrophe

although I'm not sure I've ever seen that used in preference to
U+0027.

jiffyclub:
Looks like it's a bit of a toss up:
- http://stackoverflow.com/questions/419718/html-apostrophe
- http://amp-what.com/#q=apostrophe

Anyway, I wrote this in Markdown and ran it through [redcarpet](https://github.com/vmg/redcarpet) with the `--smartypants` option, which does this sort of special character conversion.

jiffyclub:
That's essentially what `&#39;` is. I don't make these decisions, [SmartyPants](https://github.com/vmg/redcarpet#also-now-our-pants-are-much-smarter) does.

wking:
On Sun, Mar 03, 2013 at 07:28:56AM -0800, Greg Wilson wrote:

> Let's take it as is --- energy invested in single quotes is probably
> more profitably invested elsewhere :-)

Alright.

As I was doing my final pre-merge review, I spotted (formatting removed):

  I think it wouldn't be too hard to expand the lesson I'd planned to
  spend all day playing with ipythonblocks.

I think this should read: “…lesson if I planned…”.

jiffyclub:
No, that's not the meaning I intend. The sentence might read better reworded to "I think the lesson I planned could easily be expanded to spend a whole day playing with ipythonblocks."

(P.S. I already merged this, but I can easily update it.)

jiffyclub:
Reworded in 6c7cafa.

wking:
On Sun, Mar 03, 2013 at 10:52:42AM -0800, Matt Davis wrote:

> Reworded in 6c7cafa.

I understand the original wording better now ;).  I like the version
in 6c7cafa better than the original, but you could probably just drop
the “I'd planned” entirely.  Possibly worth another commit, but
probably not worth further discussion ;).

PR #335, opened 2013-03-03
Move Windows setup to a new stack using msysgit and a scientific Python distribution

ethanwhite:
I'm opening a pull request to facilitate discussion (i.e., don't merge this yet) of a (hopefully) simplified and more robust Windows stack and associated installation instructions.

The initial attempt is based on EPD and msysgit. I will also be testing Anaconda instead of EPD when I get the chance.

### DIFF
--- a/setup/windows.html
+++ b/setup/windows.html
@@ -5,117 +5,57 @@
 {% endblock file_metadata %}
 
 {% block content %}
-<p>[Tested using Windows 7 Enterprise, SP1 2009]</p>
-<p>For Windows we use Cygwin which provides for Linux/UNIX interaction but fully integrated with the Windows file-system (unlike, for example, a virtual machine).</p>
-<h2>Install Cygwin for bash, make, text editors, SQLite, Mercurial and Python</h2>
+<h1>For all boot camps</h1>
+
+<h2>Install the Anaconda CE Python Distribution</h2>
 <ul>
-<li>Download setup.exe from <a href="http://www.cygwin.com/">http://www.cygwin.com/</a></li>
-<li>Right-click setup.exe and select <strong>Run as administrator</strong></li>
-<li>Cygwin Setup window appears</li>
-<li>Click <strong>Next&gt;</strong></li>
-<li>Choose Installation Type window appears</li>
-<li>Select <strong>Install from Internet</strong></li>
-<li>Click <strong>Next&gt;</strong></li>
-<li>Choose Installation Directory window appears</li>
-<li>Click <strong>Browse...</strong> and select a Root Directory, or use the suggested default</li>
-<li>For Install For, select <strong>All Users</strong></li>
-<li>Click <strong>Next&gt;</strong></li>
-<li>Select Local Package Directory window appears</li>
-<li>Click <strong>Browse...</strong> and select a Local Package Directory, or use the suggested default</li>
-<li>Click<strong> Next&gt;</strong></li>
-<li>Select Connection Type window appears</li>
-<li>Select<strong> Direct Connection</strong></li>
-<li>Click <strong>Next&gt;</strong></li>
-<li>Choose A Download Site window appears</li>
-<li>Select a mirror site or add one closer to you. I wanted a UK one, so entered User URL: <code>http://www.mirrorservice.org</code> and clicked <strong>Add</strong></li>
-<li>Click <strong>Next&gt;</strong></li>
-<li>If you have not installed Cygwin before a Setup Alert dialog may appear, so just click <strong>OK</strong></li>
-<li>Select Packages window appears</li>
+<li>Download and Install Anaconda CE from <a href="http://continuum.io/anacondace.html">http://continuum.io/anacondace.html</a></li>
+<li>Use all of the defaults for installation <em>except</em> before pressing <strong>Finish</strong> make sure to check the <strong>Make Anaconda the default Python</strong> box.</li>
 </ul>
-<p>Now, finally, you can select the packages to install...</p>
+
+<h2>Install Msysgit for bash</h2>
 <ul>
-<li>Click <strong>Database</strong></li>
-<li>Scroll to sqlite3</li>
-<li>Click <strong>New column</strong> until it toggles to 3.7.3-1 (or, the latest version)</li>
-<li>Click <strong>Devel</strong></li>
-<li>Scroll to make</li>
-<li>Click <strong>New column</strong> until it toggles to 3.82.90.1 (or, the latest version)</li>
-<li>Scroll to mercurial</li>
-<li>Click <strong>New column</strong> until it toggles to 2.3.1-1 (or, the latest version)</li>
-<li>Click <strong>Editors</strong></li>
-<li>Pick your favourite editor...</li>
-<li>Scroll to nano</li>
-<li>Click <strong>New column</strong> until it toggles to 2.2.5-1 (or, the latest version)</li>
-<li>Scroll to vim</li>
-<li>Click <strong>New column</strong> until it toggles to 7.3.646-1 (or, the latest version)</li>
-<li>Scroll to xemacs</li>
-<li>Click <strong>New column</strong> until it toggles to 21.4.22-1 (or, the latest version)</li>
-<li>Click <strong>Python</strong></li>
-<li>Scroll to python</li>
-<li>Click <strong>New column</strong> until it toggles to 2.6.8-2 (or, the latest Python 2.x version)</li>
-<li>Click <strong>Shells</strong></li>
-<li>Scroll to bash</li>
-<li>Check <strong>New column </strong>until it toggles to 4.1.10-4 (or, the latest version)</li>
-<li>Click <strong>Next&gt;</strong></li>
-<li>Resolving Dependencies window appears</li>
-<li>Click <strong>Next&gt;</strong></li>
-<li>Wait until it's all done!</li>
+<li>Download and Install msysgit from <a href="http://code.google.com/p/msysgit/downloads/list?q=full+installer+official+git">http://code.google.com/p/msysgit/downloads/list?q=full+installer+official+git</a></li>
+<li>Use all of the default options in the installer</li>
 </ul>
-<p>Now, set up the path...</p>
+
+<h2>Install the Software Carpentry Installer</h2>
 <ul>
-<li>Find <strong>Computer</strong> on the desktop or in Windows Explorer</li>
-<li>Right Click and select <strong>Properties</strong></li>
-<li>System window appears</li>
-<li>Click <strong>Advanced system settings</strong>.</li>
-<li>Enter your local administrator password</li>
-<li>System properties window appears</li>
-<li>Click <strong>Advanced</strong> tab</li>
-<li>Click <strong>Environment Variables...</strong></li>
-<li>In the System variables list, select <strong>Path</strong></li>
-<li>Click <strong>Edit...</strong></li>
-<li>Edit System Variable dialog appears</li>
-<li>Add <code>;C:\cygwin\bin</code> to the end of the Variable value.</li>
-<li>Click <strong>OK</strong> to close Edit System Variable</li>
-<li>Click <strong>OK</strong> to close Environment Variables</li>
-<li>Click <strong>OK</strong> to close System properties</li>
+<li>Download the installer from <a href="https://raw.github.com/swcarpentry/boot-camps/master/setup/swc-windows-installer.py">https://raw.github.com/swcarpentry/boot-camps/master/setup/swc-windows-installer.py</a></li>
+<li>If the file opens directly in the browser select <strong>File -> Save Page As</strong> to download it to your computer</li>
+<li>Double click on the file to run it</li>
 </ul>
-<h2>Install Python setuptools</h2>
+
+<h1>For some boot camps</h1>
+<p>If your boot camp will be using the following software please install it by following the instructions</p>
+
+<h2>SQLite</h2>
+Only follow one set of these instructions depending on how your bootcamp will use SQLite.
+
+<h3>Using the Firefox addon</h3>
 <ul>
-<li>Visit <a href="http://pypi.python.org/pypi/setuptools">http://pypi.python.org/pypi/setuptools</a> and download the egg file corresponding to your Python version. Download a standard egg and not the Windows installer. For example, I downloaded <code>setuptools-0.6c11-py2.6.egg</code></li>
-<li><code></code>Save your egg into your Cygwin home directory, for example<br />
-<code><br />
-C:\MikeLocal\cygwin\home\mjj<br />
-</code></li>
-<li>Right-click <strong>Cygwin Terminal</strong> icon</li>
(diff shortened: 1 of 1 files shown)


jiffyclub:
First, thanks for tackling this!

The instructions I've been providing are like these: http://swcarpentry.github.com/boot-camps/2013-02-25-uwash-B/. Fairly similar to what you've outlined. Curious if Notepad++ can be called from Git Bash? We should probably explicitly say that they need to install Chrome or Firefox for the Notebook.

The real problem we run into is "how do I call x?". As you outlined in the email thread, it'd be great if we had a script that just set everything up so ipython, nosetests, and whatever else just worked via Git Bash.

ethanwhite:
>  Curious if Notepad++ can be called from Git Bash?
> This nice thing about Git Bash is that anything in the system path can be called directly. Unfortunately it looks like Notepad++ doesn't install itself in the path so that would take some additional work.

Do you think we really need a nice text editor since we're doing all of the Python in Notebooks these days?

ethanwhite:
Sorry for the lack of clarity, I meant do we need something better than nano/notepad, that has code highlighting, etc. Nano and notepad are already supported in this setup.

jiffyclub:
My software engineering unit doesn't use the notebook and it's nice for the students to have a code aware editor for that. Though I'm thinking of moving more of it to the notebook just for its broadcast capabilities.

ethanwhite:
OK, I'll put dealing with Notepad++ on the long-term list, but my feeling is that it's a lower priority and would come through in a second pull request.

ethanwhite:
Anaconda is looking less promising than EPD at the moment because it ships nose with a .bat file instead of an executable and the form of that .bat file is more difficult to replicate using BASH or Python. So, unless we're willing to give up running nosetests from the command line I think I'm stuck on using Anaconda at the moment. Suggestions for dealing with this are welcome, otherwise I'll continue to build on the current EPD based setup.

jiffyclub:
Can you make executable (from Git Bash) Python scripts on Windows?

ethanwhite:
> Can you make executable (from Git Bash) Python scripts on Windows?

Yes, in fact that's how I'm solving the IPython launching issue so that it works exactly like it does on the other OSs.

Now that you've mentioned this I see that that's all that `nosetests` is on Linux

``` python
#!/usr/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'nose==1.1.2','console_scripts','nosetests'
__requires__ = 'nose==1.1.2'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('nose==1.1.2', 'console_scripts', 'nosetests')()
    )
```

but my first naive attempts to port this to Windows have now failed.

```
Traceback (most recent call last):
  File "c:/Python27/Scripts/nosetests", line 9, in <module>
    load_entry_point('nose==1.1.2', 'console_scripts', 'nosetests')()
  File "c:\Python27\lib\site-packages\pkg_resources.py", line 337, in
_point
    return get_distribution(dist).load_entry_point(group, name)
  File "c:\Python27\lib\site-packages\pkg_resources.py", line 2282, in
y_point
    raise ImportError("Entry point %r not found" % ((group,name),))
ImportError: Entry point ('console_scripts', 'nosetests') not found
```

Thoughts? Besides not working at the moment my one concern is that this script includes a specific version number of nose which would seem to make it quite fragile.

jiffyclub:
It should be possible to use something from this page: https://nose.readthedocs.org/en/latest/api/core.html

For example, doing just `python -c "import nose; nose.core.main()"` acts pretty much as if you had called `nosetests` from the command line.

A script like this might work:

``` python
#!/usr/bin/env python
import sys
import nose

if __name__ == '__main__':
    sys.exit(nose.core.main())
```

wking:
On Tue, Mar 05, 2013 at 12:03:15PM -0800, Ethan White wrote:

> Besides not working at the moment my one concern is that this script
> includes a specific version number of nose which would seem to make
> it quite fragile.

The version-specific binding is probably to allow you to have several
versions of nose installed in parallel.  If you don't need that level
of precision, and I don't think we do, Matt's suggestion is the way to
go.

ethanwhite:
Nice. That did it.

Do either of you see an advantage one way or the other to using `python -c` bash scripts vs. the python scripts?

jiffyclub:
I would stick to Python scripts so that I wouldn't have to worry about `sys.argv` issues, but I don't know if it actually makes a difference. Plus, you know, calling Python, Python script...

wking:
On Tue, Mar 05, 2013 at 12:20:16PM -0800, Ethan White wrote:

> Do either of you see an advantage one way or the other to using
> `python -c` bash scripts vs. the python scripts?

I think MS Windows is more likely to be able to auto-launch Python
scripts (with a .py extension).  It's also one-less process if you can
avoid Bash ;).

ethanwhite:
Thoughts on where should I put nano and sqlite3?

They need to be in the path so the three obvious options are:
1. `C:\Windows\` which is in the path, isn't going anywhere, but involves putting files in a core system area
2. In Python's Scripts directory, where they don't really belong but then we're only messing with one area of the file system (since I'm already putting `nosetests` and `ipython` scripts in there) and it's an area most of them installed as part of the boot camp.
3. Somewhere custom, but that requires me to figure out how to programmatically add a location to the Windows path, which I don't know how to do.

jiffyclub:
What about a `~/.bin` or `~/bin` directory? Is the difficultly creating the custom directory or adding it to `PATH`?

jiffyclub:
If it is `PATH` you're worried about and you're using Git Bash you can just make a `.bashrc` file and do `export PATH=/my/new/path:$PATH`. We did  that successfully at UW.

ethanwhite:
Thoughts on where I should put the associated code in the SWC repos? Options:
1. Assets, under the `tools` directory
2. Bootcamps, in a new `installation` directory
3. Website, in the `setup` directory

ethanwhite:
Looking through the discussion I'm wondering if there is any reason not to merge the current pull request to change the documentation while I'm working on the tool. It's certainly better than what's on the website now.

wking:
On Wed, Mar 20, 2013 at 05:32:04AM -0700, Ethan White wrote:

> Thoughts on where I should put the associated code in the SWC repos? Options:
> 1. Assets, under the `tools` directory
> 2. Bootcamps, in a new `installation` directory
> 3. Website, in the `setup` directory

I think the current idea is to store new setup instruction tidbits in
boot-camps/gh-pages for easy mixing and matching by instructors.  The
thread where this was initially proposed was off-list (bad @gvwilson!
;), but I ended up with:

Wed, Jan 16, 2013 at 10:37:59AM -0500, W. Trevor King:

> Wed, Jan 16, 2013 at 02:55:32PM +0000, Mike Jackson:
> 
> > Could the _setup_xyz.html files be held setup/ and not bootcamps/ and/or
> > more structure provided within bootcamps/ e.g.
> > 
> > bootcamps/
> >   setup/
> >   2013/
> >    2013_NN_place.html
> >   2012/...
> >    2012_NN_place.html
> 
> In Pelican [1], the setup chunks will probably join the other bootcamp
> chunks in content/templates [2](or possibly
> content/templates/bootcamps).
> 
> > otherwise bootcamps/ will grow very short and fat over time.
> 
> I'm agnostic on the deep-vs-nested directory structure for bootcamps
> and posts.  Pelican currently only supports metadata extraction from
> the filename, but I've got a patch pending that can work on the full
> path [3].
> 
> > The setup information for a boot camp should be farmed out to a separate 
> > page, if desired, else the boot camp page could grow to be very long.
> 
> For anything too long for a short bootcamp blurb, there's also the
> boot-camps repo where you can build longer pages with a looser
> structure [4,5].
> 
> > For <em> versus <strong>, I used <strong> in the original setup 
> > instructions as I think it looks nicer, but I've no strong attachment 
> > either way.
> 
> If you're deciding based on looks, that sounds like a job for CSS ;).
> For buttons, it's possible that we should be using `<code>` or roll
> our own `<span class="gui-button">`.
> 
> > Perhaps a different heading than "Setup Not Required" for pre-installed 
> > tools in case people skim the heading, one of the OS sub-headings and then 
> > assume there's nothing to do. Something like "You already have..."?
> 
> +1 (or “You may already have…”).
> 
> Cheers,
> Trevor
> 
> [1]: https://github.com/wking/swc-website/tree/pelican
> [2]: https://github.com/wking/swc-website/tree/pelican/content/templates
> [3]: https://github.com/wking/pelican/commit/b6fd8cb60a0fc93ac6917dad11be7f8642d6a0b5
> [4]: https://github.com/swcarpentry/boot-camps/wiki/Boot-Camp-Pages
> [5]: https://github.com/swcarpentry/boot-camps/wiki/Stand-Alone-Boot-Camp-Sites

And Greg responded with:

Thu, Jan 17, 2013 at 06:39:32PM -0500, Greg Wilson:

> Hi
> 
> > > Yeah, but we're not planning to compile those pages, are we?  In which
> > > case, people are copying and pasting...
> > > They're compiled with Jekyll by GitHub, and Jekyll adds a Liquid
> > > extension to support includes [1](Liquid is similar to Jinja).
> > 
> > OK, so is that the right way to go?  Get this out of the 
> > website/bootcamps directory and put it in the generic bootcamps repo, so 
> > that people can compile it into their own home pages?

After which the discussion died ;).  This is probably something that
still needs hashing out (here or on discuss@).

ethanwhite:
Let's hold off on merging for another hour or so. I may have everything work well enough to include the installer in this pull request.

ethanwhite:
OK, here is a complete draft including the windows setup script, which is now integrated into the instructions.
1. Switches the Python distribution to Anaconda
2. Adds a script which installs nano and makes ipython work naturally
3. Places the nano files in Anacondas scripts directory. This may not be ideal but is way easier than the other options so I'd suggest that we put it in this way and then if we want we can go back and deal with the complexities of paths in Windows later.
4. Includes the setup script in the website repo. As per above it's not clear if this is best, but again I'd suggest leaving this decision as is for the moment and then moving things around later if desirable.

I have tested this on a couple of different 32-bit Windows 7 machines and will probably test run it at the bootcamp I'm running this weekend. More testing along with comments on the code and the new instructions are most welcome.

gvwilson:
Trevor, can you please try out the latest and see if it passes your tests?

wking:
On Wed, Mar 20, 2013 at 02:06:05PM -0700, Ethan White wrote:

> OK, here is a complete draft including the windows setup script,
> which is now integrated into the instructions.

This looks pretty good to me :).  A few notes on the setup script
(which I would put in boot-camps:setup/ along side my
hopefully-merged-eventually swcarpentry/boot-camps#12 ;):
- It looks like Anaconda supports Python 3.3 [1](http://docs.continuum.io/anaconda/changelog.html?highlight=python#id1), but presumably SWC
  still wants Python 2.7.  Maybe that's taken care of by “Use all of
  the defaults for installation” but I'd probably still use:
  
    #!/usr/bin/env python2.7
- You could use StringIO to avoid writing nano.zip to disk.  The docs
  say [2](http://docs.python.org/2/library/zipfile.html#zipfile.ZipFile):
  
    Open a ZIP file, where file can be either a path to a file (a
    string) or a file-like object.
- It probably wouldn't be a bad idea to add some `with` blocks [3](http://www.python.org/dev/peps/pep-0343/) or
  `try`/`except` to ensure you cleanup borked installs if something
  goes pear-shaped.
- Use `os.join()` instead of `python_scripts_directory + 'ipython'`.
- For the `ipython` wrapper, can you use the simpler:
  
  #!/usr/bin/python2.7
  from IPython.frontend.terminal.ipapp import launch_new_instance
  launch_new_instance()
  
  That's the wrapper generated by the current IPython setupbase.py.  I
  don't see why you need to handle `ipython notebook` as a special
  case.

None of these are major issues, but I think they'd tighten the script
up a bit.

I agree that the installation instructions (not the script) should get
merged ASAP, and we can re-home that if/when we move the rest of the
setup instructions.

Wed, Mar 20, 2013 at 02:20:12PM -0700, Greg Wilson:

> Trevor, can you please try out the latest and see if it passes your
> tests?

I have no access to Windows machinces for testing, but Ethan said he
tested it on Windows 7.  Ethan, did you run my
swc-installation-test-2.py?  I imagine it will work as expected, but
it's possibly it misses stuff in the Python scripts directory (if so,
I'll patch my script to fix that).

ethanwhite:
> I agree that the installation instructions (not the script) should get
> merged ASAP, and we can re-home that if/when we move the rest of the
> setup instructions.

The installation instructions now rely on the script, so they can't be split at
the moment.

I'll have to drop this again until next week now but will try to get back
to it then. Of course others should feel free to implement changes if they
have time.

wking:
On Wed, Mar 20, 2013 at 03:00:44PM -0700, Ethan White wrote:

> Of course others should feel free to implement changes if they
> have time. 

I've cleaned up the script as I suggested (although feel free to push
back on any of my choices, I've been wrong before ;).  It's now
pending as swcarpentry/boot-camps#47.

> The installation instructions now rely on the script, so they can't
> be split at the moment.

The linkage is weak, so I think they can be split.  Once
swcarpentry/boot-camps#47, we just update the URL in this PR and merge
it.  This keeps downloadable setup scripts collected in a single
location, which (I think) outweighs the overhead of maintaining the
script/instruction pair in two locations.

ethanwhite:
> I've cleaned up the script as I suggested (although feel free to push
> back on any of my choices, I've been wrong before ;). It's now
> pending as swcarpentry/boot-camps#47.

Looks great. I just gave it the thumbs up on the pull request.

> > The installation instructions now rely on the script, so they can't be split at the
> > moment.
> 
> The linkage is weak, so I think they can be split. Once
> swcarpentry/boot-camps#47, we just update the URL in this PR and merge
> it. This keeps downloadable setup scripts collected in a single
> location, which (I think) outweighs the overhead of maintaining the
> script/instruction pair in two locations.

A little miscommunication here. I just meant that the instructions don't work
without the script. So we don't want to update the instructions before the
script is in place and have folks trying to download a file that doesn't exist.
I'm perfectly happy with the split location.

Let me know what the URL will be when swcarpentry/boot-camps#47 is
merged and I'll update this PR to get rid of the script and use the new URL.

wking:
On Wed, Mar 20, 2013 at 06:26:37PM -0700, Ethan White wrote:

> Let me know what the URL will be when swcarpentry/boot-camps#47 is
> merged and I'll update this PR to get rid of the script and use the
> new URL.

https://github.com/swcarpentry/boot-camps/blob/master/setup/swc-windows-installer.py

ethanwhite:
OK, I think we're all set. Merge away unless we're still missing something. If this goes in soon we can work out any kinks using my bootcamp this weekend as a test run.

wking:
On Thu, Mar 21, 2013 at 07:12:08AM -0700, Ethan White wrote:

> Merge away unless we're still missing something.

Merged with 392f2045bef61e9ffde3e18f0be688a889dac677.

ethanwhite:
Thanks Trevor & Matt. It's been fun working with you on this.

PR #345, opened 2013-03-06
Recent work on the instructor's guide.

gvwilson:
Moving some stuff from the web chapter into an extras chapter, adjusting the Python description to match, updating build files to match.


wking:
On Tue, Mar 05, 2013 at 07:35:15PM -0800, Greg Wilson wrote:

> - After conversation with Jon Udell, removing XML stuff from web chapter:

I'm happy with the XML stuff moving to an appendix, but Google Fusion
tables?  I'm moderately horrified ;).  I also think that this would be
a good opportunity to put 'XML' in the section title since “Reading
Documents” sounds important, but learning about XML is only important
if you're being forced to deal with someone else's XML ;).

PR #346, opened 2013-03-06
Adding an option to echo paths to the badge creation script

gvwilson:


wking:
On Tue, Mar 05, 2013 at 09:11:52PM -0800, Greg Wilson wrote:

> - Adding an option to echo paths to the badge creation script

I don't think the PNG import change belongs in this commit, and I
liked it better at the top of the file rather than buried inside some
function.  If I want to only print/raise the error if I need the
module, I usually use something like:

  try:
      import some_module
  except ImportError as _some_module_error:
      some_module = None

  def some_function():
      if some_module is None:
          raise _some_module_error

gvwilson:
The problem is that the PNG module isn't installed on the software-carpentry.org server, and I was getting tired of seeing that error message every time I issued a badge.  (The PNG module isn't actually needed to _issue_ badges, only to extract information from them.)

wking:
On Wed, Mar 06, 2013 at 12:57:16PM -0800, Greg Wilson wrote:

> The problem is that the PNG module isn't installed on the
> software-carpentry.org server, and I was getting tired of seeing
> that error message every time I issued a badge.  (The PNG module
> isn't actually needed to _issue_ badges, only to extract information
> from them.)

I understand.  The idiom I posted earlier keeps the attempted import
at the top of the module (where it's conventionally located), but
saves any import errors for later.  If it turns out that you excercise
a code path that needs the module, _then_ you raise the old exception
(or print a warning).

gvwilson:
Updated.

PR #352, opened 2013-03-06
Wrote blog post about first round of LBL workshops

gvwilson:
And cleaned up some mistakes I made while trying to 'git mv' and 'git rebase' and stuff. Yes, I could clean this up, but I'm tired, and this needs to get up today.


jiffyclub:
I'd just like to know how you always end up with a bunch of previously merged commits in your branches.

wking:
On Wed, Mar 06, 2013 at 11:09:11AM -0800, Greg Wilson wrote:

> - Blog posts

Err, looks like this adds some merge-conflict markers in
`workshop-for-high-energy-physics-at-ucl-part-2.html`.

wking:
On Wed, Mar 06, 2013 at 11:49:24AM -0800, Greg Wilson wrote:

> On 2013-03-06 2:32 PM, Matt Davis wrote:
> 
> > I'd just like to know how you always end up with a bunch of previously 
> > merged commits in your branches.
> > 
> > Me too.

Possibly you're using:

  $ git checkout -b my-feature

instead of:

  $ git checkout -b my-feature origin/master

to start your branches?

wking:
On Wed, Mar 06, 2013 at 12:02:33PM -0800, Greg Wilson wrote:

> Ah --- yes.  I'm usually in the 'master' branch on my laptop, which I've 
> learned to only ever update using 'git pull upstream master' so that 
> it's in sync with swcarpentry/master,

This works if you're current master branch points to something in the
upstream master's history (in which case the pull gives you a
fast-forward merge).  If you had some work in your local master that
was not in the upstream master, the pull will try and merge upstream's
master into your local master.  It's probably better to just do

  $ git reset --hard website/master

but if you have been developing in master instead of in feature
branches, this could blow away your local development (it will still
be available via reflog).  You can also do:

  $ git pull --ff-only website

which will only work if you don't have local development in master.

> then I 'git checkout -b something-or-other' to go forward from that.
> Is that a mistake?

This will work if:
1. You actually are running it in the master branch, and
2. Your master branch doesn't have local development work in it.

You don't have to worry about satisfying these assumptions if you run:

  $ git checkout -b something-or-other website/master

jiffyclub:
I suspect that at this point your master branch and website/master don't have the same history. If you're not working on anything there it would probably be a good idea to hose your master branch and make a new copy based on website/master.

wking:
On Wed, Mar 06, 2013 at 04:54:29PM -0800, Greg Wilson wrote:

> On 2013-03-06 7:26 PM, Matt Davis wrote:
> 
> > I suspect that at this point your master branch and website/master 
> > don't have the same history. If you're not working on anything there 
> > it would probably be a good idea to hose your master branch and make a 
> > new copy based on website/master.
> > 
> > Which would be:
> > $ git checkout whatever # so that I'm not in 'master'
> > $ git branch -D master # to blow my existing 'master' away
> > $ git branch master upstream/master # to create a new 'master' on my 
> > machine based on swcarpentry:website/master

That should work, but I'd probably do:

  $ git checkout master
  $ git reset --hard upstream/master

PR #358, opened 2013-03-09
updated AUB, added KAUST bootcamp

ahmadia:
I've added the KAUST bootcamp that will be following the AUB workshop and made some small changes to the AUB site.

### DIFF
--- a/bootcamps/2013-03-aub.html
+++ b/bootcamps/2013-03-aub.html
@@ -28,7 +28,7 @@ their own laptops or connect to remote workstations during the boot camp.
   <li>how (and how much) to test programs</li>
   <li>reproducible documents and workflows with make and the IPython notebook
   </li>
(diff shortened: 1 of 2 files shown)


wking:
On Sat, Mar 09, 2013 at 07:27:29AM -0800, ahmadia wrote:

> - updated AUB, added KAUST bootcamp

There's a missing '</li>' in:

```
<li>using Python in high performance computing
```

  </ul>

In both 2013-03-aub.html and 2013-03-kaust.html.

Also, "OS X" [1](http://en.wikipedia.org/wiki/OS_X) seems to be preferred over "Mac OS" [1,2], so
"Apple's OS X" might be more accurate than "Mac operating system".

ahmadia:
Thanks, I think I addressed both issues.

PR #371, opened 2013-03-13
Added EventBrite widget to UCDavis boot camp page

amyrhoda:


gvwilson:
I think we also need the line:
    `<meta name="registration" content="open" />`
so that "open" will show up in the middle column of the bootcamps index.html page (to tell people we're accepting registration, and that it's not restricted).  Can you please update?

PR #372, opened 2013-03-13
Making it possible to install small changes more quickly by eliminating checks.

gvwilson:
As long as the server remains as slow as it's been these past few days, we need a way to publish small changes without the 10-minute delay of checking for tabs etc.


pipitone:
Looks good to me.

PR #376, opened 2013-03-15
Blog post about new boot camps

amyrhoda:
I used "cellpadding" for the table because it was too crowded. It would be better to fix it in the CSS but I didn't want to mess with CSS at this time of night and risk breaking older posts. I would also like the table to be centred, but again not worth messing with at the moment.


jiffyclub:
If you use `class="table"` in the `<table>` element you'll pick up the Bootstrap CSS for tables and it'll look good. I'll put in a quick fix.

jiffyclub:
Classes added in f6db41f.

PR #391, opened 2013-03-18
Blogging around cumulative enrolment

gvwilson:


wking:
On Sun, Mar 17, 2013 at 07:18:46PM -0700, Greg Wilson wrote:

> - Blogging around cumulative enrolment
> 
> -- File Changes --
> 
> ```
> A blog/2013/03/cumulative-enrolment.html (13)
> A files/2013/03/cumulative-enrolment.png (0)
> ```

s/enrolment/enrollment/ for US spelling?  Do we have a standardized
(-ised?) spelling for SWC?

wking:
On Mon, Mar 18, 2013 at 04:03:04AM -0700, Greg Wilson wrote:

> go ahead and merge a fix.

Done with 86b7d071db6838b8bda1be1c0af9d29e89394115.

I'm not sure if it's worth adding a redirect (somehow…), but my feed
reader already collected the entry linking to the old location.

wking:
As a side note, that plot is somewhat surprising.  With boot camps
starting at tᵢ holding sᵢ students, it looks like you're plotting:

  (tᵢ, ∑ⱼⁱsⱼ)

with linear interpolation between those points.  This leads to
diagonal rises when nobody is actually in a boot camp.  It would make
more sense (to me) to use something like gnuplot's `with steps` [1](http://www.gnuplot.info/demo/steps.html),
which:

  connects consecutive points with two line segments: the first from
  (x1,y1) to (x2,y1) and the second from (x2,y1) to (x2,y2)…  `steps`
  traces first the change in x and then the change in y.

This would give discrete bumps of sᵢ at the beginning of each boot
camp, with flat lines between boot camps.  Probably not worth
changing, but since it took me a moment to work out what was going on,
I'll post this anyway (I'm also curious about how well GitHub handles
Unicode math in emailed comments ;).

PR #394, opened 2013-03-18
Updated registration information for these two camps.

amyrhoda:
I'm not sure if there's a "full" value for the "registration" meta key and I didn't know where to look to find out, so perhaps you can tell me. :) (The AUB event is full. I put a note to that effect on the page itself but it doesn't show up in the list.)


gvwilson:
Easy enough to add a 'full' option --- I'll do that.

ahmadia:
Looks good, merging.

PR #396, opened 2013-03-18
Changed to R bootcamp

amyrhoda:


amyrhoda:
Hold on -- I just realized my local version doesn't have the link to Duke that we added yesterday. I don't know why not. Hopefully this pull request will only change the _content_r.html line.

gvwilson:
Hm --- I just did "git pull upstream master" to refresh my local version, and 2013-05-nescent.html _does_ include a link to the Duke camp, and uses _content_r.html.  2013-05-duke.html includes a link to NESCent and _content_r.html as well, so I think we're good.

amyrhoda:
That's good; thanks.

PR #400, opened 2013-03-18
Switching to bar chart

gvwilson:
As requested...


wking:
On Mon, Mar 18, 2013 at 04:59:23PM -0700, Greg Wilson wrote:

> As requested...

Actually, I suggested a stepped interpolation, not a bar chart ;).  I
think I liked the initial line graph better, and it's probably not
worth worrying about anyway :p.

gvwilson:
Then let's leave it as is.

PR #401, opened 2013-03-19
Blog post about why I write HTML and a lesson from that for structured data

gvwilson:
Comments welcome.

### DIFF
--- a/blog/2013/03/html-markdown-and-structured-data.html
+++ b/blog/2013/03/html-markdown-and-structured-data.html
@@ -1,0 +1,42 @@
+{% extends "_blog.html" %}
+{% block file_metadata %}
+<meta name="post_id" content="5418" />
+<meta name="author_id" content="wilson.g" />
+<meta name="title" content="HTML, Markdown, and Structured Data" />
+<meta name="post_date" content="2013-03-19" />
+<meta name="category" content="content" />
+<meta name="category" content="opinion" />
+{% endblock file_metadata %}
+{% block content %}
+<p>A lot of my colleagues in the Software Carpentry project use Markdown for documents, and wonder why I still mostly write in HTML. The main reason is that I like consistency, and also like to have control over typography. For example, take a look at the following snippet from a book I'm working on:</p>
+<blockquote><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Of course, it's possible to have more than one<br/>dictionary in play at any time. We <em>could</em> say:</p>
+<pre>def alt_metrics(left, right, p=1.0):
+    '''Find max synchrony between two digrams.'''
+    return left.rescale(p) * right.rescale(log(p)).T</pre>
+<p>but that's not numerically stable. What we need to do is<br/>adjust the relative proprieties after multiplying the matrices.</p></blockquote>
+<p>Did you notice that the first block of text was indented (signalling that it's the start of a new paragraph) while the second isn't (to show that it's a continuation of the same paragraph)? The text would be readable without that clue, but I find it easier to follow the flow when it's present. Right now, I lay that out as:</p>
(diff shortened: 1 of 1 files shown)


wking:
On Tue, Mar 19, 2013 at 07:36:15AM -0700, Greg Wilson wrote:

> - Blog post about why I write HTML and a lesson from that for structured data

There's a s/LateX/LaTeX/ typo, and I'd probably use `<code>` for
`CamelCaseNotation` and `pothole_case_notation`.  I didn't catch any
other editing issues.

wking:
On Tue, Mar 19, 2013 at 09:20:02AM -0700, Greg Wilson wrote:

> any comments on the content itself?

I think that the benefit of continuation paragraphs isn't worth the
boilerplate it requires ;).  Both HTML and Markdown are markup
languages, and selecting the write language for any problem is a
question of balancing access to low level stuff with ease of use /
brevity of code.  For me, writing webpage content falls on the
"brevity of code" side of the spectrum, so I use Markdown (also
because it's ikiwiki's default format [1](http://ikiwiki.info/)).  If you expect to be doing
a lot of reshuffling, etc, it makes sense to me to go with something
more formal (although I'd probably chose ReST over HTML).

HTML is the target language, but the only reason I'd chose HTML as the
source language would be if I couldn't translate from a simpler source
language with sufficient fidelity (where “simpler” is a matter of
personal preference ;).  It's fine if HTML is the simplest choice for
you, and I definately agree that more structure is better.  You just
need to balance the benefit of adding the structure against the effort
required.

beroe:
How do I view the file as HTML? I keep getting to commit pages....?

wking:
On Tue, Mar 19, 2013 at 09:55:39AM -0700, beroe wrote:

> How do I view the file as HTML? I keep getting to commit pages....?

I don't think GitHub will serve the raw page as `text/html` (and it
needs to be compiled with `compile.py` to fill in some missing pieces.

Another benefit to Markdown, etc., is that you have have a wider
audience of literate readers without an intervening rendering step ;).

jiffyclub:
That was a lot to read to find out this isn't actually an important problem...

I think setting up this idea that you need to defend your choice of HTML against Markdown fanatics is a straw man. Do I like writing content in Markdown? Yes. Do I think HTML is probably the better choice in a situation when a limited set of default formatters doesn't give you what you need? Yes. I don't think five paragraphs is necessary on the topic.

You don't give any space to the _downsides_ of your HTML choice. Namely readability. I had to download and compile the site to read your blog post. If it were written in Markdown I could have read it right on GitHub. Even reading the source of the blog post locally in my text editor is painful, not so with Markdown.

If this is really a blog post about structured vs. unstructured data the examples should use data **and** talk about the benefits/downsides of each. Going with the most structured data format you can find isn't as win-win as you make it sound here. What you gain in flexibility you pay for in portability, readability, and ease of use.

ahmadia:
I agree with Matt, I tried to read Greg's post from my own and had to
navigate over to the raw HTML file on github, which is not quite as bad as
having to download/compile (which comes from templated HTML, not HTML
itself), but I would have preferred to just read a MarkDown version.  But
indeed, I was able to read it without having to compile anything.

I think of MarkDown as a lightweight/easy interface to HTML, because you
can do most things that you want in MarkDown without having to go into
HTML.  I also appreciate the versatility in being able to handle different
output formats, although HTML does seem pretty ubiquitous these days.

I think you should work in whatever tool your colleagues can all agree on
(or that you can force them to use).  Greg eventually had to back down from
Subversion, I suspect we will eventually win the MarkDown battle as well.
 I'm not sure what hills we're leaving him to defend :)

A

khinsen:
The main conclusion from this experiment is that you have to choose: either write HTML, or use Github for requesting comments. Reading your HTML on Github is a pain and doesn't create sympathy for the content, whatever it may be.

gvwilson:
Clearly needs a top-to-bottom rethink.

Issue #293, opened 2013-02-11
Provide instructors to Oxford DTCs boot camp - 2013-05-09/10


mikej888:
- Where: University of Oxford
- Host: Jonathan Cooper (@jonc125)
- Audience: Oxford DTCs only
- Instructors: Mario Antonioletti (@marioa), Shoaib Sufi (@shoaibsufi)
- Helpers: Phil Fowler (possibly, 1 day)
- Web page: http://software-carpentry.org/bootcamps/2013-05-oxford-dtc.html 
- Registration: http://www.eventbrite.com/event/5656292122 (password-protected)
- Mailing list: http://lists.software-carpentry.org/listinfo.cgi/oxford-dtc-software-carpentry.org 
- Repository: currently, https://github.com/jonc125/boot-camps/tree/2013-05-oxford-dtc, eventually https://github.com/swcarpentry/boot-camps/tree/2013-05-oxford-dtc

mikej888:
Pinged Jonathan, Mario and Shoaib:
- Web page will shortly appear at  http://software-carpentry.org/bootcamps/index.html#future, when pull request #308 is merged.
- Asked if they want EventBrite and mailing list set up.
- Asked that they let me know when they decide if they want to use GitHub or BitBucket for a repository.
- Asked if they want uk@lists.software-carpentry.org pinged for any interested helpers.
- Asked Shoaib and Jonathan for their GitHub usernames.

jonc125:
I've drafted an initial page for the boot camp at http://jonc125.github.com/boot-camps/2013-05-09-oxford-dtc/, and will submit a pull request for it to go under the SWC account when we're ready to advertise. I've also made a (very initial) start on material in the same repository (https://github.com/jonc125/boot-camps/tree/2013-05-oxford-dtc).

I've never used EventBrite before - how easy is it to filter registrations, since this is a closed-access boot camp?

Pinging for helpers would be good - we'll have a few internal people, but it's looking like probably not enough!

amyrhoda:
Re: filtering registrations on EventBrite -- what you can do is set the event to be private and password-protected. Then no-one can sign up unless they have a password.

mikej888:
Thanks Amy. Is that OK for you Jonathan? If so I'll set it up. Likewise, do you want a mailing list?

mikej888:
Pinged the October Oxford bioinformatics boot camp attendees to see if anyone wants to help out.

jonc125:
[...]

PR #414, opened 2013-03-22
Update details for May boot camp at Oxford

jonc125:
I've replaced some of the standard (imported) text with brief content specific to our boot camp. See also #293.


gvwilson:
There appears to be a conflict between this pull request and other things that have already been merged; can you please update and we'll try again?

jonc125:
Hopefully that'll resolve the issue. Let me know if the way I've merged isn't helpful for merging back - my git experience thus far has mostly been git-svn, so I'm not very practised at less linear workflows!

PR #417, opened 2013-03-24
Thinking about using the Notebook as a teaching tool

gvwilson:


wking:
On Sun, Mar 24, 2013 at 11:53:11AM -0700, Greg Wilson wrote:

> - Thinking about using the Notebook as a teaching tool

Typo:

  “Unfortunately, the requirement for two independent screens makes
  Studio Show in most situations:”

should be

  “Unfortunately, the requirement for two independent screens makes
  Studio Show impossible in most situations:”

(added “impossible”).

wking:
In response to your “"what if?" clobbers state”, I'd like to excerpt
from PEP 20:

  Namespaces are one honking great idea -- let's do more of those!

;).  Of course, that means that you need to keep a general idea of
your existing namespace, but that shouldn't be too bad if you avoid
cluttering it up in ways that the subsequent notebook cares about.
It's too bad that things like:

  a = object()
  a.b = 5

don't work, but you can always do something like:

  class Namespace (object):
      pass
  a = Namespace()
  a.b = 5

If you put the class definition in some out of the way place (e.g. a
folded cell, once that's possible), it's easy to create a safe new
playground on the fly.

PR #421, opened 2013-03-27
Updating the how-to

gvwilson:
End-to-end edit; needs careful review from boot camp organizers, and then comparison with the checklists in the `assets` repo's `checklists` directory.


wking:
There's some trailing whitespace in f852443.

wking:
I'd use:

  “The first thing you need to decide is your intended audience.”

instead of:

  “The first thing you need to decide is who your intended audience is.”

Also:

  “Boot camps traditionally have been free,”

instead of:

  “Boot camps traditionally have been,”

wking:
I don't think:

  “(which you may choose to respect or not)”

is a constructive addition.

PR #465, opened 2013-04-08
Added install-[dev-]rsync targets to makefile to rsync local builds to website

pipitone:
Building on the webhost server is sloooow.  Here are two targets that allow you to build the website locally (and hopefully much more quickly) and then push updates to the website (either software-carpentry.org or dev.software-carpentry.org) using rsync.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ BLOG_RSS_FILE = $(OUT_DIR)/feed.xml
 # iCalendar feed
 ICALENDAR_FILE = $(OUT_DIR)/bootcamps.ics
 
+# website user@hostname
+WEBSITE_USERHOST = swcarpentry@software-carpentry.org
+
 # Standard site compilation arguments.
 COMPILE = \
 	python bin/compile.py \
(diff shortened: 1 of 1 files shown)


gvwilson:
Have you tested by building and pushing to dev.software-carpentry.org?
- G

pipitone:
Yes.

jiffyclub:
:+1:

wking:
On Mon, Apr 08, 2013 at 09:50:26AM -0700, Jon Pipitone wrote:

> - Added install-[dev-]rsync targets to makefile to rsync local builds to website

Shell quoting around arguments?  Also, I usually see $() for expanding
Make variables, although ${} works too.  Putting these together, for
example:

```
rsync -avz "$(OUT_DIR)/" "$(WEBSITE_USERHOST):software-carpentry.org/"
```

pipitone:
@wking Nice catch.  Why do you suggest putting shell quoting around arguments? 

In any case, it be done.

wking:
On Thu, Apr 18, 2013 at 11:42:02AM -0700, Jon Pipitone wrote:

> @wking Nice catch.  Why do you suggest putting shell quoting around arguments? 

Thu, Apr 18, 2013 at 11:46:54AM -0700, Greg Wilson:

> Merged #465.

I'll submit a new PR with my suggested quoting ;).

wking:
On Thu, Apr 18, 2013 at 03:24:19PM -0400, W. Trevor King wrote:

> On Thu, Apr 18, 2013 at 11:42:02AM -0700, Jon Pipitone wrote:
> 
> > @wking Nice catch.  Why do you suggest putting shell quoting
> > around arguments?
> > …
> > I'll submit a new PR with my suggested quoting ;).

Ah, nevermind, I misunderstood your initial statement.

The quoting is just to avoid problems if someone does something crazy
like:

  $ make OUT_DIR='path with spaces' install-rsync

For example, if someone was trying to build the website in a mysGit
environment on MS Windows and set `OUT_DIR=~/tmp/`, the expanded
`Documents and Settings` might be a problem (if Windows still uses
that for home directories).  Rather than figure out if spaces or other
shell-sensitive characters are likely, it's easier to just quote
everything ;).

PR #499, opened 2013-04-19
Peg JQuery to a specific version

jiffyclub:
Set JQuery version to 1.9.1 so we don't run into problems with JQuery 2.0 coming out.

Also deleting the files `jquery-1.7.js` and `swc.js` since I don't think they're used anywhere.


wking:
On Fri, Apr 19, 2013 at 10:37:23AM -0700, Matt Davis wrote:

> Set JQuery version to 1.9.1 so we don't run into problems with
> JQuery 2.0 coming out.

Yay!  [1](https://github.com/swcarpentry/website/pull/21#issuecomment-11205208).

jiffyclub:
I know, I know. :expressionless:

PR #516, opened 2013-04-30
Results from a pre-assessment

gvwilson:
Some results to chew on...

### DIFF
--- a/blog/2013/04/pre-assessment-results.html
+++ b/blog/2013/04/pre-assessment-results.html
@@ -1,0 +1,56 @@
+{% extends "_blog.html" %}
+{% block file_metadata %}
+<meta name="post_id" content="5434" />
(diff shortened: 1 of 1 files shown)


wking:
On Tue, Apr 30, 2013 at 06:25:35AM -0700, Greg Wilson wrote:

> - Results from a pre-assessment

Minor edits:
- “4 postdoc” → “4 postdocs”
- “1 general public” → “1 member of the general public” (maybe
  “layman”?)
- I'm not sure what “special student” means.
- I'd remove the quotes in the “current expertise” descriptions.
- Backticks around `results.txt`, `paper.txt`, ….

About the content, this is pretty exciting :).  I think this is why we
think teaching version control is so important.  Very few of our
students even know it exists.  It also looks like we should consider
tacking some matplotlib stuff onto the usual Python stuff.

PR #531, opened 2013-05-10
Blog post about Git vs. Subversion and why we need to collect feedback

gvwilson:
The last two paragraphs are the key...


wking:
On Fri, May 10, 2013 at 07:28:56AM -0700, Greg Wilson wrote:

> - Blog post about Git vs. Subversion and why we need to collect feedback

If you ignore Git's index, and equate the local repository with
Subversion's index, you can use the following command mappings to
reproduce your Subversion workflow:

  Subversion   Git
  add          commit -a
  commit       push
  update       pull
  diff HEAD    `fetch && diff HEAD`
  diff         diff HEAD

Only one of those is awkward.  Is the problem just dealing with
problems when things go wrong for newbie users?

I think teaching the index, local, and remote repositories as separate
concepts makes distributed development more transparent, but I like my
ideas to come in tiny, chainable chunks ;).  Will the assessment
compare uptake on Git vs. Subversion, or are we only measuring Git
(and Mercurial?) effectiveness?

On a more minor level.  `&hellip;`.  This will all be nicer in my
ideal Pelican/UTF-8 world ;).

ahmadia:
I think it's a huge mistake to introduce the git concepts associated with
remote repositories before you've taught the concepts associated with
forking/branching development.  The fundamental git model is nonlinear;
trying to teach it as a linear version control system will only frustrate
students. (Why do I have to worry about this "master' thing? Why can't I
refer to versions by number? Where did my history go? HELP!  I HAVE A
DETACHED HEAD!)

One big "ah hah" moment for me learning git was when I realized that the
process of merging two local branches is identical to the process of
merging local and "remote" work.  After that, everything is just working
with copies and making sure they're up to date.  (That is, until somebody
asks you to rebase off master for a pull request :)

Yes, there are more concepts to learn in git, but half of the battle is
getting out of the SVN anti-patterns in my opinion (and I clearly have one).

ahmadia:
Re-reading the full post (somehow my brain misfired as soon as I saw "branching").  Yes, I agree this is critical.  Good on Caitlyn for taking the interest and initiative on this.

wking:
On Fri, May 10, 2013 at 08:24:54AM -0700, ahmadia wrote:

> (That is, until somebody asks you to rebase off master for a pull
> request :)

I think `rebase -i` has a pretty helpful UI, but it does seem to be a
bit of a hurdle.  I wish we had a clearer vision of what the hurdle
was.  Is it a fixable UI problem?  Is it an (unfounded) fear of
messing with history (reflog will bail you out [1](https://www.kernel.org/pub/software/scm/git/docs/user-manual.html#reflogs))?  Is it just
frustration about cleaning up your history [2](https://www.kernel.org/pub/software/scm/git/docs/user-manual.html#patch-series)?  Rebasing conflicting
work is never much fun, but having a clean, organized history makes
figuring out why something happened so much easier.

jiffyclub:
I'd suggest rewriting this with an emphasis on the feedback aspect. Otherwise it reads a bit like an anti-Git troll post.

jiffyclub:
In a boot camp today, will have to give it some thought later.

jiffyclub:
I'd suggest this organization:
- Feedback from students is essential to Software Carpentry, otherwise how can we improve our lessons for the next round?
- For example, we've recently switched to teaching Git instead of SVN and we'd like to know if students are coming away with a worse, better, or the same understanding of version control depending on which tool we teach.
- We'll be sending out requests for feedback soon and you can help Software Carpentry tremendously by responding.

jiffyclub:
Post looks better now. It glosses over the fact that complexity isn't the only consideration when choosing which VCS we teach (How many projects use SVN? Where would I host an SVN repo during a boot camp?), but I'm okay with publishing it as is since this is a good motivator for the assessment.

PR #534, opened 2013-05-12
Updating and consolidating setup instructions

gvwilson:
By popular request, I have brought the installation instructions in setup/index.html into line with what's currently in the 'boot-camps' repo, consolidated everything into one file, and included setup instructions for R (thanks, Jenny).  Please look it over and send corrections; in particular, I'd like a better list of what our Windows installer actually installs, and instructions for installing Make on top of Git Bash if people want to do it the hard way.


ahmadia:
Two main comments:
- Can't we just leave the choice of recommending a bundle (Canopy vs. Anaconda) to the instructor and leave the recommendations neutral in this file?  I feel like we're going to have hurt feelings any other way we try to approach this.
- I don't think SQLite should be a mandatory install.  Do we have a place where we're collecting optional packages and how to install them?  Even a section that says "Extra packages.  Your instructor may ask that you also install one or more of these packages.  If so, here's how"

ahmadia:
@gvwilson - I've pushed my suggested fixes to your branch here: https://github.com/gvwilson/website/pull/4

If you accept them into your branch, the pull request here _should_ update, although I've never done this particular magic before :)

wking:
On Sun, May 12, 2013 at 08:12:26AM -0700, Greg Wilson wrote:

> By popular request, I have brought the installation instructions in
> setup/index.html into line with what's currently in the 'boot-camps'
> repo, consolidated everything into one file, and included setup
> instructions for R (thanks, Jenny).

This doesn't smell DRY, but sometimes it's worth repeating yourself if
it makes the student's life easier.  In this case, I don't really see
the benefit.

I think that having a number of small per-OS files (like we currently
have) makes things seem less overwhelming.  I'm not sure what we gain
by going to a single file.

Is the problem that some instructors don't want to bother setting up
custom per-boot-camp install?  I think a cleaner solution to that
would be to streamline the setup process, with include-able blurbs for
each package/OS pair.  The layout discussed in the unfortunately
off-list “mocking up a simpler way to handle setup instructions” back
in January sounded good.  Putting this in the boot-camps gh-pages for
Jekyll inclusion sounds like the best idea to me, and we could include
the generic suggestions in our example bootcamp [1](http://swcarpentry.github.io/boot-camps/1900-01-01-example-page/) to show folks how
it works and give anyone not interested in adding a gh-pages website
something to fall back on.

wking:
On Sun, May 12, 2013 at 11:32:47AM -0700, Greg Wilson wrote:

> Instructors have complained that the existing arrangement (one file per 
> OS) is confusing.  I toyed with a table (OS as column headings, things 
> as row titles, multicells when install instructions were the same, 
> popups when you click on the X in the cell), but it was too clever (and 
> made no sense at all when printed).

Would it still be too confusing if we dumped the “How to test” stuff
from `setup/` in favor of my script [1](https://github.com/swcarpentry/boot-camps/pull/12).  Then the main page is just
“Hi!  Click the link to your OS”.  That doesn't sound particularly
confusing to me, certainly no more confusing than a single page with
off-topic comments about other OSes.

I think that the real key for clarity is having per-bootcamp
instructions that only mention packages needed by that boot camp.
Then if folks say “the general instructions are not clear!”, they have
an easy way to roll their own (without reinventing all the hard-won
installation wisdom collected in the centralized location).

ahmadia:
> Thanks --- do Anaconda and Canopy both now include Nose

Yes.

wking:
On Sun, May 12, 2013 at 01:11:46PM -0700, Greg Wilson wrote:

> We still expect every boot camp to have the exact install
> instructions it needs on its own page (which is what we've converged
> on). This is the general-purpose fallback.

In that case, I don't have much of an opinion about what goes in here
;).

PR #535, opened 2013-05-12
Removing the long-form 'journal' version of the blog

gvwilson:
Nobody but Greg ever looked at it...


wking:
On Sun, May 12, 2013 at 01:48:40PM -0700, Greg Wilson wrote:

> Nobody but Greg ever looked at it...

That doesn't surprise me if there were no links pointing to it ;).

PR #538, opened 2013-05-13
Adding country indicators to boot camps

gvwilson:
This change puts a country flag beside each boot camp in the bootcamps/index.html page, and another one beside the boot camp title on the boot camp's page.  Almost all of the files checked in are flag icons in various sizes; other changes are to the bootcamps/_bootcamp.html template page, the index page, and the compilation script (to pick up the new metadata).
- FIXME: the rendering of the flag beside the boot camp title is ugly --- see for example build/bootcamps/2013-06-wise.html and the screenshot below.  CSS to fix this would be much appreciated.

screen shot 2013-05-13 at 10 39 06 am


gvwilson:
Added to new-look web site.

PR #539, opened 2013-05-13
Added new bibliography sections and links

mikej888:
After an e-mail exchange with Erik Bray on whether there is a collection of links to articles etc (that instructors may want to draw upon during their boot camps) I've added my set to the bibliography page under the headings "Blogs", "Hints, tips, advice and cautionary tales" (links to specific blog articles, news items, and other things), "Humor with a point" (lighter-hearted links e.g. cartoons and animations but which may make valid points).


embray:
:+1:

wking:
cf2226b looks good to me to, but I don't understand why you merged
master in ca8bcee.  Were there merge conflicts?

With my crazy editor hat on, I see an extra space before the `</a>` in
`Is it worth the time </a>`.  Also, the blogs and humor lists are all
“title - description”.  In those cases, it may be better to use a
definition list (with `<dl>`, `<dt>`, and `<dd>`).  Fix these (or not)
as you see fit ;).

mikej888:
@wking, er no. I submitted my pull request fine, then updated my repository via a pull and wrote a short unrelated blog article which I pushed. I forgot to create a new branch!

wking:
On Thu, May 16, 2013 at 09:03:51AM -0700, mikej888 wrote:

> I forgot to create a new branch!

That would do it :p.  Do you feel inclined to fix the space before
`</a>`?  How about `<dl>` vs. `<ol>`?  If you're OK with the current
situation, I can fix the `</a>` (on top of cf2226b), merge that, and
close this PR by hand.

By the way, thanks for pointing out 'Coding Confessional'.  I
especially like the limerick :p.

mikej888:
@wking, I'd fix the space but am having trouble getting a new copy of my repository - that whole "git submodule" thing is grinding everything to a halt. So please could you fix the "</a>", thank you. I'll watch my branches in future!

PR #546, opened 2013-05-16
Summarizing where we are

gvwilson:
I needed to summarize the state of play for a potential backer, and thought I'd turn it into a blog post that people could point to when asked, "What is this anyway?"

### DIFF
--- a/blog/2013/05/where-we-are.html
+++ b/blog/2013/05/where-we-are.html
@@ -1,0 +1,203 @@
+{% extends "_blog.html" %}
+{% block file_metadata %}
+<meta name="post_id" content="5443" />
+<meta name="author_id" content="wilson.g" />
+<meta name="title" content="Where We Are (More or Less)" />
+<meta name="post_date" content="2013-05-16" />
+<meta name="category" content="opinion" />
+{% endblock file_metadata %}
+{% block content %}
+<p>
+  Software Carpentry's aim is to teach scientists basic computing concepts and skills
+  so that they can get more done in less time and with less pain.
+  As context, Bob Bixby recently wrote that:
+</p>
+<blockquote>
+  Three orders of magnitude in machine speed
+  and three orders of magnitude in algorithmic speed
+  add up to six orders of magnitude in solving power.
+  A model that might have taken a year to solve 10 years ago
+  can now solve in less than 30 seconds.
+</blockquote>
+<p>
+  A million-fold speedup is pretty impressive,
+  but faster hardware and better algorithms are only two sides to the triangle.
+  The third is development time,
+  and while it has improved since 1987,
+  the speedup is measured in small percentages,
+  not orders of magnitude.
+  For most scientists,
+  getting the code to do the right thing
+  is now a bigger bottleneck than its running time.
+</p>
+<p>
+  That's where we come in.
+  We teach scientists
+  (usually grad students, since they have both need and time)
+  what they ought to know
+  before they tackle anything with "peta" or "cloud" in its name.
+  Our hope is that if we give people basic skills,
+  they'll be better able to take advantage of more sophisticated things.
+</p>
+<p>
+  Between January 2012 and July 2013,
+  over 100 volunteers will have run 92 two-day workshops for over 3000 scientists.
+  A typical two-day curriculum is:
+</p>
+<ul>
+  <li>
+    <strong>Day 1 a.m.</strong>:
+    Introduction to Unix shell.
+    We show participants a dozen basic commands,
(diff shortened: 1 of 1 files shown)


ahmadia:
This looks very good.  Thanks for the update.

wking:
On Thu, May 16, 2013 at 12:47:56PM -0700, Greg Wilson wrote:

> - Summarizing where we are

In general this looks good to me.  It overlaps with the (stalled?)
lessons-learned paper, but I don't think we can bang on this drum too
hard ;).

You have “but the "what" for scientific software is still very much up
in the air”.  Did you mean “the "why"”?  It's hard to imagine
understanding “how” without understanding “what” ;).

I think “by watching us work, people learn more than we realized…”
would read more cleanly.  I initially parsed it as “we were teaching
them by watching us work”, which is too introspective ;).

Again, you have “how” and “what”.  Maybe I'm missunderstanding what
“what” is :p.  I imagine “what” is the syllabus you list, “how” is
what the instructor does to teach it, and “why” is why the students
should show up ;).  I expect scientist instructors could more
naturally put “why” in context than non-scientist instructors (or at
least give that impression to students who were thinking of signing
up).  I don't see their angle on “how”…

`&hellip;` ;).

Links to the assessments?

In the data management paragraph, I'd replace “they” with “students”.

'nother `&hellip;`.

In the DiRAC paragraph, “want to use it” should be “want to use the
facility”, because the previous “it” referenced the driver's license.

blog/2013/05/where-we-are.html
```diff
@@ -0,0 +1,203 @@
...
+  the speedup is measured in small percentages,
+  not orders of magnitude.
+  For most scientists,
+  getting the code to do the right thing
+  is now a bigger bottleneck than its running time.
+</p>
+<p>
+  That's where we come in.
+  We teach scientists
+  (usually grad students, since they have both need and time)
```
joonro: This line seems to be missing something: they have both **need and time**

blog/2013/05/where-we-are.html
```diff
@@ -0,0 +1,203 @@
...
+    and how to use the web to do science better;
+    the hard part is to figure out what.
+    We can talk generalities about data management,
+    but they want tools and procedures.
+    As for web programming,
+    all we can show them in half a day is
+    how to create security holes...
+  </li>
+  <li>
+    We need to persuade the powers that be&mdash;the NSF,
```
npch: Should we use the more general term "research funders" rather than the specific "NSF"?

sprossiter:
"We teach scientists (usually grad students, since they have both need and time)
  what they ought to know before they tackle anything with "peta" or "cloud" in its name."

I like the style but not the substance of this sentence. The peta/cloud bit suggests good software skills are only important for big-data and big-computational-power problems. Don't we want this to be more inclusive?

blog/2013/05/where-we-are.html
```diff
@@ -0,0 +1,203 @@
...
+  For most scientists,
+  getting the code to do the right thing
+  is now a bigger bottleneck than its running time.
+</p>
+<p>
+  That's where we come in.
+  We teach scientists
+  (usually grad students, since they have both need and time)
+  what they ought to know
+  before they tackle anything with "peta" or "cloud" in its name.
```
jiffyclub: I agree with others that this is exclusive. Most researchers will _never_ tackle anything with "peta" or "cloud" in its name, but they still need us.

wking: On Fri, May 17, 2013 at 05:50:44PM -0700, Matt Davis wrote:

> Most researchers will _never_ tackle anything with "peta" or "cloud"
> in its name, but they still need us.

I tackled about a million peta-molecules of theobromine for desert an
hour ago.  For a moment, I was on cloud nine ;).

blog/2013/05/where-we-are.html
```diff
@@ -0,0 +1,203 @@
...
+  they'll be better able to take advantage of more sophisticated things.
+</p>
+<p>
+  Between January 2012 and July 2013,
+  over 100 volunteers will have run 92 two-day workshops for over 3000 scientists.
+  A typical two-day curriculum is:
+</p>
+<ul>
+  <li>
+    <strong>Day 1 a.m.</strong>:
```
jiffyclub: I'd change these to "Day 1 - Morning" and "Day 1 - Afternoon" to avoid confusion with "1 AM" and "1 PM".

jiffyclub:
What about a concluding paragraph with a few words about where we're going? Or do we know?

PR #552, opened 2013-05-17
New page for DSSG camps, added Elliot Hauser to metadata

amyrhoda:
Or would you like me to make two pages for the two dates? This way only one camp will show up on the calendar...


gvwilson:
I'm OK with just the one camp showing up in our calendar.

PR #567, opened 2013-05-26
Blogging about the future of Software Carpentry operations

gvwilson:
Comments gratefully accepted.


jiffyclub:
Ambitious! But on target. Sounds like we need a web app!

ahmadia:
I think the only bit I'm uncomfortable with is the implication that postdocs who are interested in applying for faculty positions are (or should join) our volunteer pool.  My general understanding from my own faculty search was that this was a fairly low-value piece of my c.v.  I'd be more than overjoyed to see a counter-example to this.

karthik:
Seconding @ahmadia's comment. In the current job market teaching experience counts for very little at a R1 school. Even the experience of having taught a semester long course doesn't amount to much (compared to writing more papers). Also time spent teaching software skills is seen as a distraction from real research. But I think there are still many folks interested in teaching and traveling even without this particular incentive.

gonuke:
As a (perhaps weak) rebuttal to @ahmadia & @karthikram , SWC instructor experience can be valuable for those who have no other teaching experience.  My experience is that even R1 schools would like to know that someone has at least dipped their toe in the water.  That said, is someone already has teaching/TA experience as a grad student, then I'd agree that this is less important.

jdblischak:
I agree that the actual teaching experience will not largely impact a faculty search. I would envision the benefit to the postdoc would be the free travel to ESU for networking purposes. For example, if our hypothetical postdoc Paula had met an ESU professor at a conference, she could try to grab lunch with the professor and learn more about that particular department and/or find out if she knows of any colleagues at other universities that are looking for new junior faculty specializing in their subfield. A totally different situation could be that she has a collaborator at ESU. They successfully completed one project, but the scientific relationship ended there. She could use this opportunity to get her collaborator excited about working together on a new project.

PR #568, opened 2013-05-27
Blog post on what the world might look like if we win.

gvwilson:


blog/2013/05/what-does-victory-look-like.html
```diff
@@ -0,0 +1,130 @@
...
+  The second thing is <strong>comprehensive testing</strong>.
+  This isn't just a special case of task automation,
+  although tests that aren't automated might as well not exist
+  as far as people forking a piece of work are concerned.
+  It isn't primarily about ensuring that software is doing what it's supposed to,
+  either.
+  Instead,
+  as we teach people in our boot camps,
+  the primary purpose of testing is
+  <em>to define what "correct" actually is</em>.
```
npch: And of course, for research, "correct" is generally defined as "what the community believes is currently the best hypothesis that matches the evidence".

blog/2013/05/what-does-victory-look-like.html
```diff
@@ -0,0 +1,130 @@
...
+  <li>Better algorithms are better than better hardware.</li>
+  <li>The tool shapes the hand.</li>
+</ol>
+<p>
+  If,
+  ten years from now,
+  the average scientist understands these as well as she understands significance,
+  correlation coefficients,
+  and p-values,
+  I for one will think that we've won.
```
npch, line 128: I think then that there's an eighth - the software you write is not a different class of citizen from the software you download and install on trust. You may be better or worse at writing it than others, and the software itself may be better or worse, but in the end it is all just lines of code, and anyone can be taught the skills to develop it.

Also, this might mean that we don't have the situation where scientists are afraid to ask questions to do with code. That's how we know we've won.

blog/2013/05/what-does-victory-look-like.html
```diff
@@ -0,0 +1,130 @@
...
+  new ways to measure contributions,
+  automatically tracking the provenance of every calculation&mdash;each
+  has emerged from its chrysalis and is just waiting for its wings to dry
+  so that it can take flight.
+</p>
+<p>
+  Our job is to give scientists the skills they need to nurture these ideas.
+  For the last post in this series,
+  I'd therefore like to talk about what victory will look like&mdash;i.e.,
+  how scientists lives will be different
```
ahmadia: should be scientists'

ahmadia:
Hi Greg, another well-written blog post!  I'd love a little more story about how science has evolved so far, and why we are the next logical step in science's maturity as a human enterprise.

jiffyclub:
Thank you, thank you, thank you for putting line breaks in your HTML, now I can actually read it in the PR!

Also, the post looks good.

PR #582, opened 2013-06-04
Blog post about the finances of running boot camps

gvwilson:
Careful review gratefully appreciated.


jiffyclub:
"enrollment" is misspelled in the image file name.

jiffyclub:
Nevermind, guess "enrollment" is the American spelling.

vsbuffalo:
There's likely a multiplicative effect too, as grad students that took a SC course then teach other grad students and friends what they've learned.

kaythaney:
The notion of charging for a "third instructor who doesn't show up" makes me rather uncomfortable, and I reckon would be rather off-putting to institutions / external partners (remember, these folks exist, too). Having a loose "sponsorship" model for events is a well understood one, at least in my experience, where you make sure you're covering cost but don't line-item every single cost. For a non-profit I work with in the US and UK who run 2-day events with data scientists and charities, we have a sheet where we spec out event cost, in-kind instructor and admin cost, etc., then approach sponsors on a looser basis (for our next event, we have an industry sponsor giving us two quite hefty donations that will pay for much of our operation cost as well for the year). 

I worry that framing it as such comes off as a ghost cost or a "gotcha" sort of practice that you're publicising.

npch:
Do we have strong evidence that charging a deposit or fee for learners would mean that we couldn't get waived room fees? Or only anecdotal?

benwaugh:
I have asked at UCL and they would definitely charge for rooms if we charged a fee. I didn't explicitly ask about a returnable deposit.

PR #594, opened 2013-06-09
Wise helpers

amyrhoda:
Added four of our helpers to the WiSE page. You might want to hold off on merging this because I don't have a picture for Amy Rich. Also obviously we have lots more helpers than this... Maybe I should change the header to "Some of our helpers"?


gvwilson:
That sounds good to me --- I'd be OK with just names for the helpers, 
rather than pics and bios.
- G

amyrhoda:
Still have to add Amy Rich's picture -- I'll do that this afternoon and push it this evening during my noctural Internet visit where I lurk outside the closed cafe and work in my car. Remind me to never try and work from my mom's place again...

amyrhoda:
This one is safe to merge now.

PR #595, opened 2013-06-09
Added venue and opened registration for Notre Dame camp

amyrhoda:


gvwilson:
Merging from the command line.

PR #623, opened 2013-06-26
Update website stories to reflect core boot camp material

jkitzes:
Believe it or not, I did not forget that I promised to do this! Small updates here, really - took out a few extra promises  for material that's not taught in the majority of boot camps and evened out the promises a bit across stories. Also moved Bhadra (the "web" story) to the final position, since this is the least frequently taught material.

### DIFF
--- a/about/audience.html
+++ b/about/audience.html
@@ -6,59 +6,6 @@
 {% block content %}
 
 <section>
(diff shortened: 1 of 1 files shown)


ethanwhite:
Looks good to me.

